PR #92, opened 2013-04-30
Fixes #12 with a new PR, updated to reflect pre-release and build metadata.

jeffhandley:

PR #102, opened 2013-06-03
Mark 2.0.0 as RTM (remove prerelease label)

jeffhandley:
This is based on the changes from #92 (which provides the summary).


ratchanonsuttawas:
> This is based on the changes from #92 (which provides the summary).
>

PR #115, opened 2013-06-14
No leading zeroes on numeric identifiers

isaacs:
Closes mojombo/semver#112


dol:
This change is in 2.0, but not part of the EBNF definition at https://github.com/semver/semver/blob/master/semver.md#backusnaur-form-grammar-for-valid-semver-versions .
It would be nice to clarify this part in the EBNF.

Issue #104, opened 2013-06-06
SemVer badge/shield.


jmalloc:
Just a quick thought I had - I think it would be beneficial to have a small, official "SemVer" icon/badge/shield to display on websites, README files, etc, that indicates a project uses semantic versioning.  It's often hard to tell from tag names alone, and this would remove any doubt.

PR #126, opened 2013-06-24
Add first run at a SemVer Badge

EddieGarmon:
For #104 

Comments? Suggestions?


jwdonahue:
Where is this "badge" meant to be used?

EddieGarmon:
> Where is this "badge" meant to be used?

Please see #104 linked above

alexandrtovmach:

bhean:
I like the idea, but the SemVer specification use "Semantic Versioning" as well, so I think should be necessary notice WHICH version of SemVer are you using in your project. According with the "classic" (and very nice) badge style, I mean something similar to: 

img

KuryKat:
And you can even make it better if you use the logo Query Param

img

Markdown code:
`img`

buhtz:
I would prefer using such a badge file from upstream repo instead of a tracker like shields.io.

ljharb:
that wouldn’t support the dynamism that shields.io does.

buhtz:
> that wouldn’t support the dynamism that shields.io does.

What kind of dynamism?

PR #127, opened 2013-07-01
Added description for pre-stable public API versioning

ChrisWren:
When reading the description for major version zero I came across the line "Anything may change 
at any time." I would like to create a more clear explanation of how versioning should take place before the first stable version of the public API so that developers can have meaning behind their pre-stable versioning.

This is especially an issue in node.js development as by default saved modules will pick up new patch versions but not new minor versions. If developers don't properly version their modules in a pre-stable state they can potentially break when users pick up new patch versions.

To define this pre-stable state I have proposed moving the definition for major and minor changes to the right, meaning that API-breaking changes result in a minor version being incremented and both backwards-compatible functional additions as well as bug fixes result in a patch version being incremented.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -67,8 +67,15 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development. Anything may change
-at any time. The public API should not be considered stable.
+1. Major version zero (0.y.z) is for initial development. The public API should not be considered stable.
+
+1. Major version zero's minor version (0.Y.z) MUST be incremented if any backwards
+incompatible changes are introduced to the public API. Patch version MUST be reset to 0 when minor
+version is incremented.
+
+1. Major version zero's patch version (0.y.Z) MUST be incremented if new, backwards
+compatible functionality is introduced to the public API or backwards
+compatible bug fixes are introduced to the public API.
 
 1. Version 1.0.0 defines the public API. The way in which the version number
 is incremented after this release is dependent on this public API and how it
--- a/semver.md
+++ b/semver.md
@@ -171,11 +171,6 @@ them.
 FAQ
 ---
 
-### How should I deal with revisions in the 0.y.z initial development phase?
-
-The simplest thing to do is start your initial development release at 0.1.0
-and then increment the minor version for each subsequent release.
-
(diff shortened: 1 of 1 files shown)


jwdonahue:
Current SemVer has two ways to be "prerelease", the 0.x.x series and anything with a -prerelease tag on it.  Why two?  Well it's just a fact of history at this point.  If we ever get to a 3.x.x series in SemVer, I hope we close that loop-hole permanently.  I see nothing wrong with starting out at 1.0.0-unstable, but then we have the problem where -unstable has higher precedence than -beta or -prerelease and that's just another reason why I have been tempted to propose yet another standard.  

Basically, the problem that @ChrisWren is trying to solve here, stems directly from the 0.x.x loop-hole. Require the prerelease tag and remove the loop-hole, the problem goes away, you have the full semantics we are looking for.  Here's a proposed workaround that I am pretty sure the current standard allows:

0.0.1-dev       // Start of sprint, unpublished except for commits.
1.0.0-a.ring.0 // Internal dog-food release.
1.0.1-a.ring.1 // Internal + partners dog-food release.
1.0.2-alpha    // First public release.
1.0.3-beta     // Second public release.
1.0.4-prerelease1    // Third public release.
1.0.5    // First "stable" release.

Notice how many times I had to use the word "release" to describe what happened? SemVer seems to imply tighter constraints on that word than most developers actually use in practice.  The 1.0.0-a.ring.0 version is a published prerelease.  Even the 0.0.1-dev version was "released" by the developer to the rest of the team the instant they committed their change.  Unfortunately, the word "published" is not mentioned in the spec, despite the fact that it appears to be what was actually meant by its use of "released".

We have a stable spec at 2.0.0, with fairly good adoption at this point and while I do feel there's some work to be done within that scope, I think there's good reason to move on to 3.x.x discussions.  I believe I just offered an acceptable workaround for @ChrisWren, that has a high probability of not breaking anything and without changing the 2.0.0 spec.  If somebody cries foul over the missing 1.0.0 (no tags) version, who cares?

jwdonahue:
#221

jwdonahue:
@ChrisWren, please close this issue at your earliest possible convenience.  Thank you.

alexandrtovmach:

adamralph:
Is there any reason the maintainers have not closed this PR? There are also several other very old ones with clear indications that they are not going to be accepted.

alexandrtovmach:
@adamralph Thank you for paying attention to that, I'll go through staled things a bit later

lastthyme (approved):

jwdonahue:
Linking to a more recent and related PR and discussion: #915 #923

PR #152, opened 2013-08-12
Making it clear how 0.X.Y versions work

krainboltgreene:
I've come up against a lot of libraries that stick to 0.X.Y versions and when challenged claim "oh we might change the API". SEMVER is supposed to give you the freedom to change your public API as long as you document the severity of that change in the version. This line gives developers something to hide behind to avoid the scary 1.X.Y declaration when in reality they're well past a stable release and being depended on by others.

I find this to be both annoying and detrimental to the use of Semantic Versioning. One major point of SEMVER is that *APIs can safely change as long as a corresponding change happens to their Version*.

As such I've removed the wording "api can change" because it's not descriptive of 0.X.Y. 0.X.Y should be simple: If you're still working on minimum viable functionality then you're still 0.X.Y. As soon as someone depends on you, then you should document your public API and bump the major version to 1.X.Y.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -69,7 +69,7 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
 1. Major version zero (0.y.z) is for initial development and SHOULD NOT be 
-considered for public consumtion. Avoid distribution to avoid dependency
+considered for public consumption. Avoid distribution to avoid dependency
 on an unstable API.
 
 1. Version 1.0.0 defines the public API. The way in which the version number
--- a/semver.md
+++ b/semver.md
@@ -68,9 +68,9 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development and shouldn't be 
+1. Major version zero (0.y.z) is for initial development and SHOULD NOT be 
 considered for public consumtion. Avoid distribution to avoid dependency
-on unstable API.
+on an unstable API.
 
 1. Version 1.0.0 defines the public API. The way in which the version number
 is incremented after this release is dependent on this public API and how it
--- a/semver.md
+++ b/semver.md
@@ -68,8 +68,9 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development. Anything may change
-at any time. The public API should not be considered stable.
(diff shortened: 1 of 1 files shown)


RWOverdijk:
Is this going somewhere? Because I agree with the point being made. I don't think that 0.x.y deserves having its own set of rules unless it literally means "don't rely on this code too much, it's not done yet".

EddieGarmon:
Maybe for 0.x.y versions: s/unstable/unfinished

krainboltgreene:
0.X.Y is the worst concept ever, and gives so many developers a scapegoat
for the fear of real releases. The excuse I see time and time again is "but
my library isn't finished" (rarely does a library ever "finish", that's the
reason we have version numbers) and "but my api could change at any time"
(that's literally what major versions are designed to document).

Frankly, any tool is broken if it tries to document the public API
compatibility for a period of life in a codebase where there are no plans
to be compatible.

Want to not be broken? Strike the definition from the record, tell everyone
to get some courage and realize their library is probably used in
production and needs to be 1.0.0 immediately.

FichteFoll:
Imo 0.x.y is needed for rapidly changing products or those which don't really have a defined public API yet. If you are in early development and following semver strictly but started with 1.0.0, you are likely approaching 100.0.0 quickly.

The only problem is that people often miss the point where their product should actually be 1.0.0 already, and you need to remind them of it. I'm all for an international "1.0.0 day" or "major 1 day" where developers revisit their old code to check if a 1.0.0 release is feasable or necessary.

What semver should do in this regard is probably reword the 0.x.y section a bit or add a sentence that clarifies how sub-1.0.0 versions *must be depended on explicitly*. Usually you would only specify `>=x.y.z <x+1.0.0` as a version selector, where x.y.z contains the latest feature addition or bugfix you need. For 0.x.y releases you MUST NOT use selector ranges but may only depend on an exact version, as anything can change any time.

krainboltgreene:
All projects can be rapidly changing. The whole point of SEMVER is
documenting the result of change and releases.

crazedsanity:
I agree with @FichteFoll in that 0.x.y releases MUST NOT use selector ranges but may only depend on an exact version, as anything can change any time.  It is both logical and follows unspoken principles of many existing libraries.

RWOverdijk:
So if I understand correctly, you're saying that 0.x.y (or any tag for that matter) is production ready, but has an unstable API, so you must depend on an exact version to make sure you don't shoot yourself in the foot.

I feel there's something wrong with this. Being able to trust bug fixes and security patches in PATCH bumps is a major win for developers. Now I have to go through change logs for every 0.x.y lib I use, if I wish to update. That should only happen for major bumps imo.

Also, @FichteFoll I don't think that's true. If you're in early development, usually there's a preview of some sort (edge). Even then, I don't think it really matters if you're on 100.0.0. It probably just means you tagged to early, or you don't know what you're doing. That's what PRs are for.

milesrout:
0.x.y is absolutely fine as it is. Versioning something doesn't mean it is finalized. It seems a lot of the comments here are just people venting that npm contains a lot of 0.x.y packages, which is something wrong with them, not with semver.

krainboltgreene:
Yeah, it's just a problem endemic of at least two huge package managers
that use SEMVER. It's totally not SEMVER's fault. Just coincidence.

I'm closing the ticket since it's like talking to brick walls.

PR #172, opened 2014-01-10
Include Maven SNAPSHOT pre-release example

leif81:
The Maven build system uses the keyword "SNAPSHOT" to indicate pre-release builds. Their definition and use fits nicely with the semver pre-release definition so it would be useful to be shown here as an example. Worth mentioning is that the "SNAPSHOT" naming convention is sometimes used by non-Maven projects too.

"you will often see the SNAPSHOT designator in a version, which indicates that a project is in a state of development"
http://maven.apache.org/guides/getting-started

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -100,7 +100,7 @@ precedence than the associated normal version. A pre-release version
 indicates that the version is unstable and might not satisfy the
 intended compatibility requirements as denoted by its associated
 normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7,
-1.0.0-x.7.z.92.
+1.0.0-x.7.z.92, 1.0.0-SNAPSHOT.
 
 1. Build metadata MAY be denoted by appending a plus sign and a series of dot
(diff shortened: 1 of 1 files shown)


leif81:
@jwdonahue Can you clarify why you believe it's not compatible?

jwdonahue:
@leif81 , -SNAPSHOT has higher precedence than -alpha, -ALPHA or -BETA, but the semantics of it should be lower, because -SNAPSHOT is a place holder for a time-stamp.  The standard doesn't need to list every popular prerelease tag in existence.  The current set seems sufficient.

alexandrtovmach:

jwdonahue:
@leif81 Please refer to [clause 3](https://semver.org/spec/v2.0.0.html#spec-item-3) in the spec:

> 3. Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version.

The -SNAPSHOT tag is applied to multiple published releases, therefore any x.y.z-SNAPSHOT version may not be unique in the world.  Definitely not SemVer compliant.

PR #206, opened 2014-07-03
Switched from Lexical Sort to Natural Sort

tedivm:
This change better first with people’s expectations. This way
1.0.0-beta2 has a lower precedence than 1.0.0-beta11, because “2” is
smaller than “11”- a point missed by lexical sorting. It also solves
the alphabetical issue  because natural sorting is essentially case
insensitive (it may or may not make sense to specify that two versions
of the same exact name but different cases is frowned upon, in which
case it’s case sensitive for our purposes).

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -95,12 +95,13 @@ version is incremented.
 series of dot separated identifiers immediately following the patch
 version. Identifiers MUST comprise only ASCII alphanumerics and hyphen
 [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST
-NOT include leading zeroes. Pre-release versions have a lower
-precedence than the associated normal version. A pre-release version
-indicates that the version is unstable and might not satisfy the
-intended compatibility requirements as denoted by its associated
-normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7,
-1.0.0-x.7.z.92.
+NOT include leading zeroes. If an identifier ends with numerals then those
+numerals MUST be considered as a separate numeric identifier. Pre-release
+versions have a lower precedence than the associated normal version. A
+pre-release version indicates that the version is unstable and might not
+satisfy the intended compatibility requirements as denoted by its associated
+normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-alpha1, 1.0.0-rc2
+1.0.0-0.3.7, 1.0.0-x.7.z.92.
 
 1. Build metadata MAY be denoted by appending a plus sign and a series of dot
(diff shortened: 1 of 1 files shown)


dumblob:
>I don't think this change should be accepted. Implementations get harder and this introduces more edge cases. How would `1.0.0-beta2a` compare to `1.0.0-beta11b`? For those that want to have versions within prereleases, we already have support for it using dot-separated segments: `1.0.0-beta.2` and `1.0.0-beta.11`. No need to add another way to do what can already be done.

@jeffhandley there are two things not to be mixed together - identifier of the pre-release in a group/type of pre-releases (betas, alphas, etc.) and then a version of the particular pre-release. In `1.0.0-beta2a` it's mixed (i.e. the group/type of pre-releases is called `-beta`, the identifier of the particular pre-release is `2` and the version of this particular pre-release is `a`).

For versioning of any (pre-/post-)release shall Semver provide **exactly one** unified way (i.e. X.Y.Z), so `a` or `b` is not acceptable. There is though an extremely easy, programmer-friendly, and intuitive at the same time solution - see [sane unified approach](https://github.com/mojombo/semver/issues/163#issuecomment-192662245 ). So implementations can be extremely straightforward.

jwdonahue:
@tedivm, this change is not required. Just add a dot to end of the non-numeric identifier (-beta.2). Already has good semantics defined in the spec.  Unless you pursue this further with additional changes to your proposal or stronger arguments in favor it, please close this PR.  If it was going to be accepted, it would have been years ago.

tedivm:
@jwdonahue - I'm not sure how you missed this but I closed this PR on October 13, 2016, well over a year ago.

tedivm:
Hmm, @jwdonahue, looking through this project it looks like you aren't a contributor. For the record it is extremely rude for someone who has no authority (or even a single contribution) on a project to go around telling people to close their pull requests. Leave that to the people who either submitted the PR or who manage the project.

jwdonahue:
> I'm not sure how you missed this but I closed this PR on October 13, 2016,

@tedivm, apologies, I don't know how I missed that either. Probably wandered in following a link from an old issue in semver/semver.

As to your other comments, I do have an open PR and I have put in about 30 hours going through the back-log and getting folks to close old issues that were either answered or weren't likely to be acted on (see SemVer #413). I think I've helped to clarify some things along the way. There's more than one way to contribute to a project than checking-in code. As for authority, unless I am banned from the project, I have all the swizzles I need to ask folks to close out old issues.

> Leave that to the people who either submitted the PR or who manage the project.

No offense, but I suspect they have day jobs. Semver/semver had 150 open issues when I started working the problem last week-end.  Got it down to 134 so far and I am not getting a lot of push back.  All I want to do is help.

PR #234, opened 2014-11-18
Address "v" prefixing in the FAQ

beck:
As suggested here: https://github.com/mojombo/semver/issues/204#issuecomment-47609174

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -311,6 +311,15 @@ No, but use good judgment. A 255 character version string is probably overkill,
 for example. Also, specific systems may impose their own limits on the size of
 the string.
 
+### Is "v1.2.3" a semantic version?
+
+No, "v1.2.3" is not a semantic version. However, prefixing a semantic version
+with a "v" is a common way (in English) to indicate it is a version number.
+Abbreviating "version" as "v" is often seen with version control. Example:
+`git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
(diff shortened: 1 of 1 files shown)


beck:
:bell:

PowerKiKi:
I would rather discourage the used of "v" prefix. The suggested text, while it clarify some aspects, leave the decision up to the reader. I would rather add a sentence to the effect of:

    SemVer discourage the use of any prefix for the sake of clarity and ease of interoperability.

beck:
Usually a good quality for interoperability (aka compatibility) is great.  What's even better is backward-compatiblity.

From semver 1.0:

> When tagging releases in a version control system, the tag for a version MUST be "vX.Y.Z" e.g. "v3.1.0".

I'm not sure why this was removed. This pull would not be needed if Version 2 addressed the issue with a "MAY" or "MUST NOT".

FichteFoll:
The current spec only specifies the version itself, not how you use or embed it. Imo it's better to make usage of semver as permissive as possible while also being strict on the actual semver part. Thus, only mentioning the `v` prefix in the FAQ and not in the spec itself makes more sense than more or less arbitrarily have a special case, because there are no special cases.

julmot:
:+1: a clearification is necessary regardless if it will be in the spec of FAQ.

beck:
@mojombo What's up with this repo? Is it dead?

Haacked:
Looks good to me.

PR #251, opened 2015-04-24
Changed MUST to MAY for resetting Minor and Patch when incrementing Majo...

deanrather:
There is no need to reset the Minor version number when incrementing the Major, as the new Major version may include the same functionality as the previous Major version.

Patch version numbers need only be reset when the functionality changes. I.e. when the Minor Version is incremented.

Resetting the Minor version to Zero looses the Minor Version Number's helpfulness in seeing at-a-glance how many new feature versions have been developed.

This makes the following legal:

```
1.2.3 -> 2.2.3     // A compatibility-breaking change was introduced, but no patches or new functionality
1.2.3 -> 2.2.4     // A compatibility-breaking patch was introduced, but no new functionality
1.2.3 -> 2.3.0     // A compatibility-breaking feature was introduced
1.2.3 -> 2.0.0     // Any of the above combinations, but you can still reset both if you like!
```

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -88,7 +88,7 @@ MUST be reset to 0 when minor version is incremented.
 
 1. Major version X (X.y.z | X > 0) MUST be incremented if any backwards
 incompatible changes are introduced to the public API. It MAY also include minor
-and patch level changes. Patch and minor version MUST be reset to 0 when major
+and patch level changes. Patch and minor version MAY be reset to 0 when major
 version is incremented.
 
 1. A pre-release version MAY be denoted by appending a hyphen and a


EddieGarmon:
I think this is excessive granularity, all that matters is a breaking change that requires human intervention, or something that should just work.

FichteFoll:
> the new Major version may include the same functionality as the previous Major version

How would that be possible? You increase major because there is a breaking change *in functionality*.

Other than that, I'm rather neutral on this. I mean, it's optional. While you can convey a bit more information regarding functionality if you want to carry over minor version numbers between multiple major versions (preferrably all), it doesn't really seem necessary to me either. Additionally, you usually increase minor if there is new functionality, but what do you do if you remove some? Reset minor to 0? Then all your information that you built up over the last few major and minor releases is lost anyway.

rlidwka:
> How would that be possible? You increase major because there is a breaking change in functionality.

You can potentially increase major version because of one bugfix that people depend on, effectively keeping functionality the same.

deanrather:
> How would that be possible? You increase major because there is a breaking change in functionality.

If you renamed a public API, for example. It's not new functionality but it is no longer backwards compatible.

j6k4m8:
I'm in agreement with allowing 'resets' — consider this case, of a user updating from one version (skipping a few) to the latest:

<kbd>1.2.3</kbd>→<kbd>2.2.4</kbd> 
<kbd>1.2.3</kbd>→<kbd>2.0.1</kbd>

Similar-*ish*, but the latter makes it clear that one patch has been made *since the roll to 2.0.0*. I think this *adds* information, since it tells you, using nothing but the latest version alone, what sorts of changes have been made since a major release. (The former instead says, "how many patches have I *ever* made" which seems not as meaningful to me.)

weakish:
Another possibility is increasing patch number for every patch and minor number for every feature.

https://medium.com/@tjholowaychuk/semver-bccee96cc9ac

sgryphon:
@deanrather "Minor Version Number's helpfulness in seeing at-a-glance how many new feature versions have been developed."

There is no requirement in SemVer that minor version does this, i.e. SemVer does not specify that minor version means this.

A minor version increment of 1 (e.g. from 1.2.0 to 1.3.0) could be 0, 1 or 7 new features, and still be compliant with SemVer. It can also increase my more (e.g. from 1.2.0 to 1.37.0) and also still be compliant with SemVer.

If you are relying on minor version being the number of features, then it is not.

jwdonahue:
You can encode feature counts in the build tag.  You are free to define build tag semantics as you see fit.

jwdonahue:
@deanrather, the correct way interpret the version string 2.1.3 is "the third patch level of the first minor version of major version 2". In other words, patch is an epoch within the minor epoch of the major epoch. Bug, feature and breaking change counts have nothing to do with it. You bump the appropriate level when you publish, according to the semantic rules and always reset the lower fields.

Please cancel/close this PR, thank you.

deanrather:
I agree.

lloydcody:
In all the time elapsed since this issue was first raised I’m yet to come across a compelling argument for *enforcing* the reset of minor and patch numbers on major number increments.

The _usefulness_ of persisting the minor and patch version numbers is a separate conversation, which many of the comments in this thread relate to.

This request is to simply make it the implementer’s choice. If they feel it would be beneficial to them to keep the number information in tact, let them do so by way of it being optional.

To that aforementioned separate conversation...
An “API” is conceptually a significantly different concern to the application logic. It delivers structured data versus delivering behaviour and logical derivations. As these separate conceptual concerns are intrinsically mapped to the major number and minor-plus-patch number respectively, we _should_ be well within our rights to release updates to either without the one impacting the other.

Note the same point cannot be (and isn’t being) made for minor and patch number resets as they share the same conceptual concern.

If I have version 1.0.0 make its way up to version 1.15.3 and then change the value type of a field in my API from a mandatory enum of [Red, Green, Blue, Orange] to [Red, Purple, Yellow] and even modify any (now illegal) row values to conform, I’ve introduced an API-breaking change without any behavioural, derivation or look-and-feel changes at all. As a user eager to trial all the goodies in my favourite Colour Picker software, I’d know immediately not to get excited by looking at a version number bump of 1.15.3 to 2.15.3. However, if I open up version 2.0.0 fully expecting it to be brimming with exciting new features, I’m going to be sorely disappointed, closing the application with feelings of despair and betrayal. All because the database guy streamlined some colours and the company practiced SemVer.

FWIW, I personally hold what I believe must be a very human and potentially historically-biased opinion of resetting the numbers. But I still can and will. However, putting our OCD and warm fuzzy feelings aside, surely SemVer MUST not enforce the loss of what some users would consider meaningful data?

To quote SemVer’s mantra: “Remember, Semantic Versioning is all about conveying meaning by how the version number changes. If these changes are important to your users, use the version number to inform them.”


DISCLAIMER: the OP created this PR to make his point that I’m not allowed to keep our software minor and patch number increments when our API was modified unless the SemVer spec itself changed. I was outraged as how else would our boss measure how hard the dev team had been working if we had to wipe out all our “progress” with one piddly API release?!


Years on and to this day, I stand by this PR being fully valid. It’s one character less too so even saves on bandwidth for everyone!

FichteFoll:
Semver is not meant to track progress on the code base for managers or even users. It is solely meant to convey the severity of changes to the API, not how many changes there have been. As such, there is no reason to change the reset of minor and patch numbers.

Instead, resetting the numbers can tell a message of how long the current major or minor branch is likely to receive updates until a new breaking change is added, although this information is quite vague and a release schedule would be far better suited.

PR #177, opened 2014-02-04
Fixes #176 by clarifying that versions are case insensitive

jeffhandley:

PR #276, opened 2015-10-03
Versions are not case sensitive

### LINKED ISSUE #176: Spec doesn't indicate whether pre-release labels and build metadata are case sensitive
Let's hope they aren't.  But can we get some clarification as to whether or not they are meant to be?

Haacked:
Fixes #176
Continued from #177

Note that even though this is technically a breaking change, _every_ change is a breaking change. The question is really how likely is this to break people.

The chances that people rely on capitalization for version precedence seems highly unlikely.

Also, SemVer applies to libraries, not necessarily specifications. Though I would agree that a library based on SemVer would probably want to increment the major version possibly putting it out of sync with the version of this spec.

Software is messy, isn't it?

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -1,4 +1,4 @@
-Semantic Versioning 2.0.0
+Semantic Versioning 2.0.1
 ==============================
 
 Summary
@@ -122,11 +122,11 @@ for two pre-release versions with the same major, minor, and patch version MUST
 be determined by comparing each dot separated identifier from left to right
 until a difference is found as follows: identifiers consisting of only digits
 are compared numerically and identifiers with letters or hyphens are compared
-lexically in ASCII sort order. Numeric identifiers always have lower precedence
-than non-numeric identifiers. A larger set of pre-release fields has a higher
-precedence than a smaller set, if all of the preceding identifiers are equal.
-Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta <
-1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
+lexically in ASCII sort order but without case sensitivity. Numeric identifiers
+always have lower precedence than non-numeric identifiers. A larger set of
+pre-release fields has a higher precedence than a smaller set, if all of the
+preceding identifiers are equal. Example: 1.0.0-alpha < 1.0.0-Alpha.1 <
+1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-BETA.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
 
 Backus–Naur Form Grammar for Valid SemVer Versions
 --------------------------------------------------
@@ -310,6 +310,13 @@ that users can smoothly transition to the new API.
 No, but use good judgment. A 255 character version string is probably overkill,
 for example. Also, specific systems may impose their own limits on the size of
 the string.
+
+### Are SemVer pre-release labels and build metadata case-sensitive?
+
+No, they are not case-sensitive. In practice, standardizing on lower-case
+pre-release labels and build metadata can be beneficial, especially when
+cross-platform scenarios are considered. Systems must be case-insensitive when
+comparing and ordering pre-release labels and when comparing build metadata.
 
 About
 -----
--- a/semver.md
+++ b/semver.md
@@ -1,4 +1,4 @@
-Semantic Versioning 2.0.1
+Semantic Versioning 2.1.0
 ==============================
 
 Summary
--- a/semver.md
+++ b/semver.md
@@ -121,12 +121,12 @@ lower precedence than a normal version. Example: 1.0.0-alpha < 1.0.0. Precedence
 for two pre-release versions with the same major, minor, and patch version MUST
 be determined by comparing each dot separated identifier from left to right
 until a difference is found as follows: identifiers consisting of only digits
-are compared numerically and identifiers with letters or hyphens are compared
-lexically in ASCII sort order but without case sensitivity. Numeric identifiers
-always have lower precedence than non-numeric identifiers. A larger set of
-pre-release fields has a higher precedence than a smaller set, if all of the
-preceding identifiers are equal. Example: 1.0.0-alpha < 1.0.0-Alpha.1 <
-1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-BETA.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
+are compared numerically and identifiers with letters or hyphens are compared in
+a case-insensitive lexical ASCII sort order. Numeric identifiers always have
+lower precedence than non-numeric identifiers. A larger set of pre-release
+fields has a higher precedence than a smaller set, if all of the preceding
+identifiers are equal. Example: 1.0.0-alpha < 1.0.0-Alpha.1 < 1.0.0-alpha.beta <
+1.0.0-beta < 1.0.0-BETA.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
 
 Backus–Naur Form Grammar for Valid SemVer Versions
 --------------------------------------------------


FichteFoll:
> every change is a breaking change

No, unless the previous breakage in case of an addition was part of the spec in the first place, in which case semver would be absolutely useless anyway. For example, If we were to add another field to semver, indicated by a new character that was not allowed previously.

Haacked:
I should have clarified. Every change has the potential to break someone because people sometimes do ill-advised things like rely on private reflection to call internals. Just read [Raymond Chen](http://blogs.msdn.com/b/oldnewthing/archive/2003/12/24/45779.aspx) for some crazy stories. :smile:

In this particular case, I'm going to argue that SemVer always intended comparison to be case-insensitive and we're effectively fixing a mistake in the spec by clarifying the intended behavior. For example, I know that NuGet implements comparison as case insensitive.

How have others implemented this in code? If most implementations do treat it as case sensitive, I can reconsider and increment the major version. Does anyone strongly feel we should go one way or another?

cbandy:
Can you please describe the motivation for this change? What are the benefits of not ordering pre-release identifiers by their ASCII code?

Haacked:
The behavior is surprising. Which version has precedent? 1.0.0-alpha or 1.0.0-Alpha?

Also, some package managers rely on the file name including the version meaning the scheme doesn't work on file systems that are case insensitive.

Haacked:
Or a better example of the surprising behavior is `1.0.0-alpha` vs `1.0.0-America`. I might expect alpha to come first, but ASCII ordering puts America first.

I'm curious how NPM handles this. We should definitely ping them. The most active contributor in the last month is @zkat so I hope you don't mind me bugging you with this question. :smile:

Haacked:
An alternate proposal is to say labels MUST be lowercase which is probably what people do in practice. For older SemVer packages, they'd be treated as lowercase by newer parsers for back compat.

Another alternate proposal is to say SemVer labels SHOULD be lowercase to prevent confusion.

zkat:
In npm, we only match exact semver versions for pre-releases when installing. If you have a prerelease suffix, it's automatically taken out of the ordering. (`satisfies` in `node-semver`). We also just [use timestamps for ordering](https://github.com/npm/npm/blob/master/lib/fetch-package-metadata.js#L117-L139).

Haacked:
Thanks for chiming in @zkat, and I apologize for my naivete when it comes to NPM. I really should spend more time with it given it's the most popular package manager out there! :smile:

I have a clarifying question. According to http://semver.npmjs.com/ it seems like you can specify version ranges that are not exact matches for pre-releases. For example, I entered `jquery` and noticed that `>=3.0.0-aaaa` matches the package `3.0.0-alpha1` but `>=3.0.0-alpha2` does not as I would expect. However, if you're not intimately familiar with the spec, it's a little surprising that `>=3.0.0-AZ` matches the same package but `>=3.0.0-az` does not.

This seems to answer my original question to you. If that site is a correct implementation, it appears that NPM is case sensitive and does the correct ASCII ordering. In practice, I would bet that nearly all NPM packages use lower-case for tags, but I have no way of determining that for sure.

I'm adding @isaacs to the discussion since we've talked in the past and you're the maintainer for npm/node-semver. My question to the both of you, if we changed the spec to be case sensitive, would that be a problem for you? I'm trying to follow the principle of lease surprise and the fact that `>=3.0.0-AZ` matches `3.0.0-alpha1` but `>=3.0.0-az` does not is surprising. It might not be surprising to you, but there are large number of developers who are used to case insensitive file systems, programming languages, etc and I'd like to make this less surprising all around if we can.

Alternatively, we could just require that pre-release tags be lowercase if we knew that 99.9% of existing NPM packages follow that rule and it wouldn't be too disruptive.

Would love to hear your opinions on this.

isaacs:
It is a bit weird, I guess.  But as it currently stands node-semver is case-sensitive:

```
> s.satisfies('1.0.0-Alpha', '>1.0.0-AZ')
true
> s.satisfies('1.0.0-Alpha', '>1.0.0-aZ')
false
```

In practice, people only use `-prerelease` versions very rarely, and they're typically very specific to a particular module's dev process.  If you're using prereleases, you probably are already in a conversation with the devs, debugging something, testing out new features, etc.  So this is rarely a problem.

I agree that it's a good idea for the spec to be clear on this point, though.  Whether semver is case sensitive by design, or case insensitive by design, it'd be an improvement in clarity.

Since node-semver is case-sensitive currently, it'd mean a change if the spec updates to be case-insensitive, but I don't think it'll be particularly controversial.

Haacked:
> Since node-semver is case-sensitive currently, it'd mean a change if the spec updates to be case-insensitive, but I don't think it'll be particularly controversial.

Thanks for chiming in @isaacs! Yeah, if we change this, we'll bump the major version of SemVer so it would be part of SemVer 3.0. The main thing I'm looking for is how controversial would this change be. I think NPM is the largest community it would affect and hearing that it probably wouldn't be controversial is good to hear.

I did talk to @mojombo about it and it is case sensitive by design at the moment. But I think we all agree that cases like the one you mentioned are odd and surprising. My guess is, very few packages actually run into this problem which is why this isn't an urgent change, but one that's nice to clear up for the long run.

I'll wait to see if anyone else has any strong objections before I fix this up and merge it.

semver.md
```diff
@@ -121,12 +121,12 @@ lower precedence than a normal version. Example: 1.0.0-alpha < 1.0.0. Precedence
...
 be determined by comparing each dot separated identifier from left to right
 until a difference is found as follows: identifiers consisting of only digits
-are compared numerically and identifiers with letters or hyphens are compared
-lexically in ASCII sort order. Numeric identifiers always have lower precedence
-than non-numeric identifiers. A larger set of pre-release fields has a higher
-precedence than a smaller set, if all of the preceding identifiers are equal.
-Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta <
-1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
+are compared numerically and identifiers with letters or hyphens are compared in
+a case-insensitive lexical ASCII sort order. Numeric identifiers always have
```
cbandy: I think the typical way to describe this is to say the identifiers are "converted to lowercase then compared".

isaacs:
As I slept on this, it occurred to me that "converted to lowercase" or "compared in a case-insensitive manner" isn't *exactly* specific in a world of many character sets and i18n's.  The turkish-i comes to mind, especially.

In order to not be a cross-domain footgun, I think maybe the specific algorithm for lower casing should be specified?  It'd probably also be worthwhile to track down an i18n expert to weigh in on the likely challenges.

crazedsanity:
I don't know if this has been specified, but it seems to me that we could limit this by recommending/specifying a limited character set.  Latin (?) `a-z0-9` would be sufficient; other languages have all kinds of quirks that probably aren't handled by semver, like text direction. 

In other words, I don't think it would be a good idea to have to worry about internationalizing a semantic version. Keep it simple.  Programming is hard enough. ;)

isaacs:
Oh, right, doesn't the spec already say that prerelease tags are [a-zA-Z0-9]?  If so, we're probably fine.

Haacked:
> Oh, right, doesn't the spec already say that prerelease tags are [a-zA-Z0-9]? If so, we're probably fine.

Yep, it does.

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]

So we're good there.

thallgren:
I'm currently fixing some issues regarding how prerelease tags are handled by the Puppet module manager, which uses the [Semantic Puppet](/puppetlabs/semantic_puppet) ruby gem. I came across the issue of case sensitivity and found my way to this PR. Is this still being considered? It would be very helpful if a conclusion was made and this PR was either merged or abandoned.

isaacs:
"Case insensitivity" is tricky, and a breaking change.

I recommend nixing this PR, and instead specifying that prerelease tags _are_ case-sensitive, and that characters are compared according to their ASCII values, since that's already what's implemented.

thallgren:
FWIW, I second the opinion of rejecting this PR. Instead, make the spec more explicit about how things work today and let the spec version stay at 2.0.0.

FichteFoll:
@thallgren, no. The spec absolutely must be updated, otherwise it would be violating itself. This would be fine as a patch release, however.

Saying  "how things work today" is kind of imprecise though, since there appear to be package managers comparing insensitively already (allegedly [NuGet](https://github.com/mojombo/semver/pull/276#issuecomment-145193945)). That said, I agree with explicitly making comparisons sensitive and referring to their ASCII code point.

isaacs:
Yes, any change to the spec requires at least a patch bump.  So, 2.0.1.

Haacked:
Ok, I'm fine with closing this. I'm not seeing an overwhelming response and changing the spec just for one implementation doesn't seem like a good use of our time.

As @isaacs says, Implementors gonna implement .

PR #282, opened 2015-11-17
Make notice about leading zeros clearer

zudov:
It was very unclear to me that "must not contain leading zeroes" is referring to "non-negative integers" rather than to "A normal version number".

First (correct) interpretation would disallow 1.01.1, second (incorrect) would disallow 0.1.0.

I tried to make it more explicit here.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -61,8 +61,8 @@ could be declared in the code itself or exist strictly in documentation.
 However it is done, it should be precise and comprehensive.
 
 1. A normal version number MUST take the form X.Y.Z where X, Y, and Z are
-non-negative integers, and MUST NOT contain leading zeroes. X is the
-major version, Y is the minor version, and Z is the patch version.
+non-negative integers. Elements (X, Y and Z) MUST NOT contain leading zeroes.
+X is the major version, Y is the minor version, and Z is the patch version.
 Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 
 1. Once a versioned package has been released, the contents of that version


FichteFoll:
Never had a problem with that, nor heard of anyone who did.

zudov:
Well, perhaps it's just me

crazedsanity:
Honestly, I'm not really sure that the change clarifies anything.

jinahya:
Besides the contents of this PR, I also support the point of improving the meaning of `leading zeros`.

crazedsanity:
After re-reading, I agree that this change is a bit easier to understand.  Disregard my last.
+1

zudov:
Of course, I don't think my current edit is ideal, and I am pretty sure someone could suggest further improvements.

semver.md
```diff
@@ -61,8 +61,8 @@ could be declared in the code itself or exist strictly in documentation.
 However it is done, it should be precise and comprehensive.
 
 1. A normal version number MUST take the form X.Y.Z where X, Y, and Z are
-non-negative integers, and MUST NOT contain leading zeroes. X is the
-major version, Y is the minor version, and Z is the patch version.
+non-negative integers. Elements (X, Y and Z) MUST NOT contain leading zeroes.
+X is the major version, Y is the minor version, and Z is the patch version.
```
jwdonahue: How about just changing it from the "...MUST NOT contain leading zeroes..." in the original as:

...except for a single character ('0') place holder, MUST NOT contain leading zeroes.

jwdonahue (commented):
Just one possible alternate suggestion for consideration.

semver.md
FichteFoll: This is implied. A single `0` number does not have leading zeroes since the single zero digit is required here. This is universally understood in all other occasions I've seen this being discussed in one way or another.

jwdonahue (approved):
Non-breaking incremental change, looks good to me.

alexandrtovmach:

PR #283, opened 2015-11-19
Add in-page anchors for direct linking

NewAlexandria:
Solution to issue https://github.com/mojombo/semver/issues/222

I would have added related CSS, but I don't know how the semver.org website is generated.  Feedback welcome on that.

### Demo

[direct link to spec \#5](https://github.com/NewAlexandria/semver/blob/master/semver.md#semver-spec-5)

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -14,8 +14,10 @@ Given a version number MAJOR.MINOR.PATCH, increment the:
 Additional labels for pre-release and build metadata are available as extensions
 to the MAJOR.MINOR.PATCH format.
 
+<a name="intro"></a>
 # Introduction
 ------------
+<sup>[[link](#intro)]</sup>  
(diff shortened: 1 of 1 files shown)


Haacked:
The website for semver is hosted at https://github.com/mojombo/semver.org. It's a Jekyll site and I believe some of the markdown processors can add header anchors automatically. We should make the change there and not in the markdown spec.

NewAlexandria:
@Haacked  just FYI, this markdown style is the same that 'we' use in [the ruby style guide](https://github.com/bbatsov/ruby-style-guide).  I'll look for any processor directive that can do the same

adjohnson916:
Relates to https://github.com/mojombo/semver.org/issues/27.

Issue #291, opened 2016-01-27
Pre-release and build meta examples contain invalid characters


clphillips:
The spec (version 2.0.0) states that:

>  Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]

Which applies to both **pre-release** (section 9) and **build meta** (section 10). **However**, the examples given include:

> 1.0.0-alpha.1, 1.0.0-0.3.7 ... 1.0.0-beta+exp.sha.5114f85

These clearly violate the rule as they contain full stops (.), which are not alphanumerics or hyphens (e.g. they fall outside [0-9A-Za-z-]).

FichteFoll:
Start reading the two sections at the beginning.

clphillips:
> MAY be denoted by appending a [hyphen|plus sign] and a series of **dot separated identifiers** immediately following the patch version

I swear it wasn't there before. ^_^

Issue #249, opened 2015-03-26
Incorrect Example 10


idmean:
Point 10 of the specification says (emphasis added):

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise ***only ASCII alphanumerics and hyphen [0-9A-Za-z-]***. Identifiers MUST NOT be empty. 

Then this example follows:

     1.0.0-beta+exp.sha.5114f85

However the identifier contains periods which are not included in `[0-9A-Za-z-]` nor are they alphanumerics or hyphens.

So is this example incorrect?

crazedsanity:
The given example is correct, as it refers explicitely to the identifiers, which are *separated* by the dots.  So `exp` is valid, `sha` is valid, and `5114f85` is valid as individual identifers; the full metadata is then `exp.sha.5114f85`.  Hopefully that makes sense.

Issue #230, opened 2014-11-01
Build metadata spec is inccoesitent with examples.


omeid:
> 1. Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. **Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]**. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+**exp.sha.5114f85**.

Either, preferably  `.` needs to be added to the allowed list of identifiers, or the example to be corrected.

PR #296, opened 2016-03-10
Slightly clarify the build metadata section

### LINKED ISSUE #293: Example for build metadata does not comply with the spec itself
Section 10, about build metadata says:

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.

Note that a period is *not* allowed in the build metadata according to this, but the last example does use periods in the build metadata (exp.sha.5114f85). This makes the section ambigious and confusing: Are periods allowed or not?

matthijskooijman:
Multiple people (#291, #249, #230, #293) have missed the fact that the
build metadata can contain multiple _dot-separated_ identifers, and were
confused enough to report a bug about the dots present in one of the
examples. This attempts to make it more obvious that multiple
identifiers might be used.

This closes #293.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -91,9 +91,11 @@ incompatible changes are introduced to the public API. It MAY also include minor
 and patch level changes. Patch and minor version MUST be reset to 0 when major
 version is incremented.
 
-1. A pre-release version MAY be denoted by appending a hyphen and a
-series of dot separated identifiers immediately following the patch
-version. Identifiers MUST comprise only ASCII alphanumerics and hyphens
+1. A pre-release version MAY be denoted by appending a hyphen followed
+by one or more identifiers. The hyphen immediately follows the patch
+version and is followed by the first identifier. If multiple identifiers
+are used, they are separated with dots.
+Identifiers MUST comprise only ASCII alphanumerics and hyphens
 [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST
 NOT include leading zeroes. Pre-release versions have a lower
 precedence than the associated normal version. A pre-release version
--- a/semver.md
+++ b/semver.md
@@ -102,8 +102,10 @@ intended compatibility requirements as denoted by its associated
 normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7,
 1.0.0-x.7.z.92, 1.0.0-x-y-z.--.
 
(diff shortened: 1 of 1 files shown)


semver.md
```diff
@@ -102,8 +102,10 @@ intended compatibility requirements as denoted by its associated
 normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7,
 1.0.0-x.7.z.92.
 
-1. Build metadata MAY be denoted by appending a plus sign and a series of dot
```
Haacked: So my concern here is that the language used to define build identifiers was pretty much exactly the same as the language used above to describe pre-releases. Why is it more confusing for build identifiers? Should we change the description of pre-release too?

matthijskooijman: Excellent point. To be honest, I wasn't interested in pre-releases, so I skipped that section when reading the spec. It makes sense to keep them consistent, so I'll update my PR.

matthijskooijman:
Commit added to update the pre-release description as well. If you'd rather have them merged into a single commit, let me know.

jwdonahue:
I think this is a non-breaking change that is mildly better than the original.

@Haacked, what's holding up acceptance of this change?

semver.md
```diff
@@ -102,8 +104,10 @@ intended compatibility requirements as denoted by its associated
...
-separated identifiers immediately following the patch or pre-release version.
+1. Build metadata MAY be denoted by appending a plus sign followed by one
+or more identifiers. The plus immediately follows the patch or
```
FichteFoll, line 108: Should also say "plus sign" like in the line above.

matthijskooijman: Indeed, pushed a fix.

matthijskooijman:
I rebased this on top of master (there was a minor conflict) and fixed the comment by @FichteFoll (I amended it into the first commit, since I was rebased anyway).

alexandrtovmach:

alexandrtovmach:
@haacked @indirect what should we do with this PR? It seems a good patch?

Issue #315, opened 2016-06-13
Should rule #9 be upgraded to a SHOULD


BurtHarris:
Currently, rule 9 indicates that use of pre-release suffixes is optional, by using the word MAY.   

In real-world implementations, like npm, consumers often depend on publishers following rule 9.   It seems to me like the rule should be updated to use the word SHOULD in place of the first occurance of "MAY".

Haacked:
Hmm, I can see how that is a bit confusing. The way I read it, it's "MAY" because not every version is a pre-release version. So a pre-release version MAY be denoted.

BurtHarris:
OK, good point.   So perhaps rule nine should begin with "**Any** pre-release version **SHOULD** be denoted by appending a hyphen..."

FichteFoll:
The way you phrased it requires "MUST" to be used.

crazedsanity:
I agree that pre-releases must be distinguished from regular/non-pre-release versions.  Having "MAY" there seems indicative of the possibility that pre-releases might not be distinguishable as such.  So rule 9 would then change to:

> A pre-release version **MUST** be denoted by appending a hyphen...

Just wanted to (hopefully) clarify the wording to ensure we're all on the same page.

BurtHarris:
I'd be happy enough with either version.

tdpsk:
I agree that in practice, pre-release tags should be mandatory because users rely on them in the context of package repositories. Therefore I have created a corresponding pull request in #316, feel free to comment and discuss.

FichteFoll:
>pre-release tags should be mandatory

This could be misunderstood, but the wording in the PR is clear.

PR #316, opened 2016-06-19
Make pre-release tags mandatory as per discussion in #315

tdpsk:


krzysiekpiasecki:
There must be only zero or one component 'prerelease' in a version number. Word ```Any prerelease``` is not precise as ```A prerelease```.

BurtHarris:
Changing this from a MAY to a MUST seems like a breaking change and might require a major version change in the Semver spec.   I believe SHOULD is a better verb, making it a non-breaking change that simply clarifies this as a best practice rather than an optional practice.
   
I'm not sure what the previous comment about zero or one component prerelease adds.      Nor do I understand why "Any" is less clear than "A", but another wording making it even clearer this is a best practice would be to start Rule 9 with "All", i.e.
```
9.  All pre-release versions SHOULD be denoted as such by by appending a hyphen and a 
...
```

krzysiekpiasecki:
@BurtHarris 

```MUST``` is totally imperative; there is no escape from this rule. So if there is a ```pre-release``` component it MUST be denoted with a leading ```hypen``` and we're not talking about changes but we`re just using a ```pre-release``` definition language.

```All``` has multiplication meaning. We're specify single version in documentation as a prototype for all of them so we're using ```MAJOR version``` not ```MAJOR versions``` and so on.

That's my point of view.

**Comment by tdpsk** [comment]
While I generally believe "MUST" is the way to go, I see @BurtHarris point about backwards compatibility. From my point of view this change should not be made in a breaking way, therefore I have adapted the pull request to state "SHOULD" instead of "MUST".

For reference, an extract from RFC 2119 which defines the keyword SHOULD: 
```
This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course. 
```

**Comment by BurtHarris** [comment]
That's now good by me.   As context, my thinking about this was stimulated (but not guided) by [this discussion at stackoverflow](http://stackoverflow.com/questions/36712067/what-module-version-should-be-used-in-the-development-branch).

**Comment by nainumsunti** [comment]
Mmm

**Comment by jwdonahue** [comment]
What does it even mean to have a prerelease version that doesn't have a prerelease tag on it?  In practice, I think the prerelease tag is slapped onto just about everything, whether it ships externally or not, until the QA team allows its removal.  The current proposed change to "Any pre-release version SHOULD..." seems like a very good and non-controversial change.  I really think that making it "Any prerelease version MUST..." is a good candidate for the first 3.0.0 version of SemVer.

@Haacked, what is holding this up?

File: semver.md [hunk 1]
```diff
@@ -91,7 +91,7 @@ incompatible changes are introduced to the public API. It MAY also include minor
 and patch level changes. Patch and minor version MUST be reset to 0 when major
 version is incremented.
 
-1. A pre-release version MAY be denoted by appending a hyphen and a
```
Haacked: Would it be clearer to simply change this to:

> A pre-release version is denoted by...

jwdonahue: @Haacked, _Is_ is not defined in [RFC 2119](http://tools.ietf.org/html/rfc2119). Standards should use standardized semantics to define themselves, yes?

BurtHarris: I've been away from this for quite a bit, but trying to re-enter this discussion in a meaningful way. 

Outside the world of this spec there is a de facto convention some pre-release packages use: a major version of "0"!    I believe this should be discouraged, rather than prohibited, as there's no way to change the past.   That's why I suggested rule #9 be updated to the "SHOULD" in the first place.   

I hope this addresses @jwdonahue's question.

alexandrtovmach:

alexandrtovmach:
Ooops 😨 
Seems like fork doesn't exist anymore, so I can't reopen it. @tdpsk if you still would like to contribute, feel free to create a new PR

ljharb:
You don't have to close and reopen to kick CI, you can click through to the individual service and rerun it.

alexandrtovmach:
@ljharb Could you provide more details how can I do that? Thanks

ljharb:
It depends on the CI service tbh - but for old PRs, it's also better to rebase them and force push to the PR branch :-)

PR #371, opened 2017-05-16
Update "Z" and "Y" examples to match "X"

amesbury:


arcticicestudio:
I think you misinterpreted definition. The `x > 0` is related to rule *4*.
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

amesbury:
*SIGH*  This is why I should never have a commit bit.  :-(

arcticicestudio:
That is not right, every help is useful in the open source development. Everyone can misunderstand some lines of code or documentation which is absolutely no problem. Opening issues and pushing pull requests doen't cost anything 😉

PR #393, opened 2017-09-07
Change MUST reset to SHOULD reset

sgryphon:
As this is relaxing a constraint, it would be a minor update (2.1) to the specification - all existing implementations would remain compliant.

The relaxation of the increment by 1 requirements from SemVer 1.0 allowed, where it was necessary, to skip an increment, e.g. from release 2.4.10, change to 2.6.0 (or 4.0.0).

But, as the specification currently stands it would not be possible change from 2.4.10 to 2.5.1 - it MUST be 2.5.0 (or 2.6.0, etc).

Changing from MUST to SHOULD still encourages the reset, but allows for situations where it is not followed.

Example, moving from version 2.4.10, there could be an extensive period of 3.0.0-beta.1, 3.0.0-beta.2, etc, that never actually gets released, but is widely known; if the decision is made to skip a version, then this change would allow 3.1.0 to be developed (releasing 3.1.0-beta.1, etc) and eventually released (as 3.1.0). Without this change, the only option would be to bump to 4.0.0.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -83,11 +83,11 @@ compatible functionality is introduced to the public API. It MUST be
 incremented if any public API functionality is marked as deprecated. It MAY be
 incremented if substantial new functionality or improvements are introduced
 within the private code. It MAY include patch level changes. Patch version
-MUST be reset to 0 when minor version is incremented.
+SHOULD be reset to 0 when minor version is incremented.
 
 1. Major version X (X.y.z | X > 0) MUST be incremented if any backward
 incompatible changes are introduced to the public API. It MAY also include minor
-and patch level changes. Patch and minor versions MUST be reset to 0 when major
+and patch level changes. Patch and minor versions SHOULD be reset to 0 when major
 version is incremented.
 
 1. A pre-release version MAY be denoted by appending a hyphen and a
@@ -115,30 +115,30 @@ have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700,
       minor, patch and pre-release identifiers in that order (build metadata
       does not figure into precedence).
 
-   1. Precedence is determined by the first difference when comparing each of
+   2. Precedence is determined by the first difference when comparing each of
       these identifiers from left to right as follows: major, minor, and patch
       versions are always compared numerically.
 
       Example: 1.0.0 < 2.0.0 < 2.1.0 < 2.1.1.
 
-   1. When major, minor, and patch are equal, a pre-release version has lower
+   3. When major, minor, and patch are equal, a pre-release version has lower
       precedence than a normal version:
 
       Example: 1.0.0-alpha < 1.0.0.
 
-   1. Precedence for two pre-release versions with the same major, minor, and
+   4. Precedence for two pre-release versions with the same major, minor, and
       patch version MUST be determined by comparing each dot separated identifier
       from left to right until a difference is found as follows:
 
       1. Identifiers consisting of only digits are compared numerically.
 
-      1. Identifiers with letters or hyphens are compared lexically in ASCII
+      2. Identifiers with letters or hyphens are compared lexically in ASCII
          sort order.
 
-      1. Numeric identifiers always have lower precedence than non-numeric
+      3. Numeric identifiers always have lower precedence than non-numeric
          identifiers.
 
-      1. A larger set of pre-release fields has a higher precedence than a
+      4. A larger set of pre-release fields has a higher precedence than a
          smaller set, if all of the preceding identifiers are equal.
 
       Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 


sgryphon:
Also note that allowing this does not change the semantic interpretation or usage of versions, which is the aim of the standard.

e.g. if Firetruck requires Ladder 3.1.0, then Ladder 3.2.1 can be correctly interpreted as being compatible, even if 3.2.0 was never released (not currently allowed by spec), the same that 3.3.0 can be correctly interpreted as being compatible even if 3.2.0 was never released (which is allowed under the current specification).

jwdonahue:
@sgryphon, what is the problem with this scenario?

master
2.3.10
2.5.0-a.dev.internal  // Start new sprint for major release.
3.0.0-alpha           // Dies in ring zero flight testing.
3.0.1-alpha           // Ships to external early adoptors.
3.0.2-beta            // Ships.
3.1.3-beta            // Ships, marketing insisted on one late feature.
3.1.4-beta            // Ships.
3.1.5-release.candidate1 // Does not ship (test lab caught fire!).
3.1.6-release.candidate2 // Ships, most customers love it.
3.1.7-release.candidate3 // Ship fix for your big partner's special corner case.
3.1.1                    // Released.
3.1.1-a.dev.internal. // Start new sprint to pay back technical debt.
etc.

While there is a prerelease tag on the version string, any point release can break anything, the interface/package is unstable. Incrementing the patch number for all internal/external builds helps retain provenance tracking back to the 2.4.10 release and provides strong ordering, leaving considerable leeway in what you put into your prerelease tags, but given the above prerelease tags, you could just as easily leave the triple pinned at 3.0.0 for the entire sprint.

I don't see how anything in the spec precludes the above and it achieves what I think you are after without modifying the spec.  Do you see any problems with it?

FichteFoll:
Could you elaborate on the part that causes "3.0.0" to never be released? Is this a marketing question?

You may pull releases in case they are erroneous, in which case a version hole would ensure, but I don't see any reason to skip a version altogether.

jwdonahue:
@FichteFoll, I can list a few, yes, but I am sure the number of reasons I don't know why, might as well be infinitely larger.  For all intensive purposes, a version 3.0.0 that was released and then removed before anyone downloaded it, is essentially a missing version.  SemVer constrains the number of dev-ops processes that lead to a correct implementation of the standard, to a smaller infinity than non-adherence to the standard, but it should not, in my view constrain that good set anymore than is absolutely essential.  Requiring any x.0.0 release is not practical, especially in light of the admission that any version can be withdrawn at any time.

The primary reason I think the above scenario MUST be allowed, is that the semantics of words used in the prerelease tags does not always sort lexicographically the way that alpha and beta just happen to do, and prefixing additional numeric or alpha fields to achieve this, just adds cognitive load to human consumers.  You can find instances in my issues responses and other places where I have probably used the -a.dev tag in order to insure that all *dev prereleases sort below -alpha.  The other work around for this limitation is to always bump something in the SemVer triple, which is the primary point of the example scenario.

Basically, all "prereleases" should have a prerelease tag on them, and if you intend to use full semantic versioning through the entire prerelease phase in order to give early adopters and testers a clue, then you could very well run right past 2.x.x and 3.0.x on your way to the 3.3.1 release, while actually publishing prerelease versions of them.  Yes, marketing will have some influence on the final decision of which version to actually "release", but so will QA and probably division leads and CEO's as well.

jwdonahue:
Prerelease is a misnomer in the standard that comes from common practices.  We publish/release a version of something under the caveat that it is not the promised release that is yet to arrive.  Maybe we hope that 3.0.0-Final.Beta will result in a 3.0.0 release, but we fail to do so do to marketing forces that suddenly require us to adhere to some recently adopted standard and we wind up releasing 3.1.x instead.

jwdonahue:
Even if never releasing an X.0.0 is a violation of a strict reading of the standard, I am not aware of any case where failing to do so, breaks anything, therefore it is a restriction I can easily ignore.

sgryphon:
@jwdonahue I guess pre-release would be a way to conform to the exact spec, and there is no way for anyone to ever know that those missing releases didn't exist.

According to the spec, the minor version MUST be reset. This could be applied internally to a non-released version, so the following sequence is valid:

2.201712.10 // Released
3.0.0-internal // Internal release with minor version reset, to satisfy the specification
3.201801.0 // Released; the actual release number you want, for some unknown reason

But the following would be technically non-conforming (but does not break any semantics):

2.201712.10 // Released
3.201801.0 // Released

i.e. you nver produced an internal pre-release with 3.0.0, although there is no way for anyone to ever know this.

Changing the wording from MUST to SHOULD would simply make this official. I don't think it would break anything.

jwdonahue:
@sgryphon , what tooling are you using for packaging and publishing?

grv87:
I don't see in the current version of SemVer any notice that `release` should be public (non-internal). So, situation described by @sgryphon is perfectly legal now, no need to change anything.

jwdonahue:
@sgryphon, take a look at [VersionMeta](https://versionmeta.org/) and [VersionSchema](http://versionschema.org/). Please let me know what you think.

jwdonahue (commented):
For whatever it's worth, I vote to reject this change. It's the wrong solution for @sgryphon stated problem. If anything, we need to make it clearer what it means to "release" something and the fact that the a publisher is under no obligation to officially release any X.0.0 version. Any X.y.z is fine, as long as X > 0, eventually.

KyNorthstar (commented):
The spirit of resetting to 0 is what you're talking about. It means that you mustn't go from `2.4.10` to `3.4.10`, or `2.4.10` to `2.5.10`. You're describing someone following the proper guidance as described today, so no change is necessary.

alexandrtovmach:

patrron (approved):

Nidacoke (commented):
How can I do everyone any idea

sgryphon:
Going through my old PRs and found that although this had been approved it never merged.

> @sgryphon, what is the problem with this scenario?
> ...
> 3.0.2-beta // Ships.
> 3.1.3-beta // Ships, marketing insisted on one late feature.
> ...
> I don't see how anything in the spec precludes the above and it achieves what I think you are after without modifying the spec. Do you see any problems with it?

Because the sequence you described just contravened the specification. Part 7 says "Patch version MUST be reset to 0 when minor version is incremented."

The sequence you describes increased the minor version but did not reset patch to 0, so does not conform to SemVer. (If you do not meet a MUST condition, then you don't conform).

Now for all reasonable purposes this is irrelevant; the sequence is perfectly fine and doesn't break any of the compatibility or relative comparisons; it is just a nit pick that it technically doesn't conform to the spec.

And yes, you could solve it by some claiming an internal version in between, whether an actual build or just internal mental gymnastics:

3.0.2-beta // Ships.
3.1.0-dummy // Comply with spec
3.1.3-beta // Ships, marketing insisted on one late feature.

But why?

If intention, and commonly accepted practice, is that it is perfectly fine (even if uncommon) to jump from 3.0.2-beta to 3.1.3-beta (and I see no reason why not), then what not reflect that in the spec and have "Patch version SHOULD be reset to 0 when minor version is incremented."

If you intend the requirement to be SHOULD (not MUST), then write it that way.

Explorer09:
Why the heck is this suggestion never accepted into the official Semantic Versioning spec?

Anyway, my two cents: I would vote for changing the keyword to MAY rather than MUST or SHOULD.

Why? Because I can think of many good reasons NOT to reset the minor or patch version once the major version field is bumped:

* The software could have "patch" versions reflecting the date of patching/release. E.g. 20260214. For such versioning it's inappropriate to reset the "patch" version field to 0 even after a major or minor version bump.
* For software that implements a particular protocol or spec, sometime I can use the "minor" version field to indicate the spec version my software was conforming to. (If the protocol or spec introduces a breaking change forcing me to make a breaking change to my API, then my API can bump both major and minor version numbers to indicate such semantics.) This versioning scheme can still follow Semantic Versioning in all other ways.

So, the MUST keyword there makes the Semantic Versioning much stricter than necessary.

jwdonahue:
@sgryphon 

> If you intend the requirement to be SHOULD (not MUST), then write it that way.

All evidence suggests that the intent was MUST, not SHOULD. That "doesn't matter if you don't" statement, is derived from the fact that even if you follow the release pattern to the letter, not all extant version histories of the product, that are publicly available on relevant package stores, will consistently reflect the pattern, due to the potential for releases being removed. In fact, not all stores will reflect the same histories.

Release history != publicly available version list. No consumer tools, I am aware of, will break, if a publishing tool allows that violation of the spec. 

My recollection of discussions from the beta through 1.0.0 version of the spec, is that @mojombo, did not intend SemVer to be a tuple of category counters.  #9 was the first challenge to the reset rule, that is preserved here, but his response prior to closing that issue, indicates that the spec "was designed to fit in well with established versioning paradigms".

Changing this rule would violate the semantics that patch is the Nth version of Major.Minor and minor is the Nth feature of Major. If you really want a triple or other tuple of category counters, you are free to create a spec and reference implementation.

Explorer09:
> 
> Changing this rule would violate the semantics that patch is the Nth version of Major.Minor and minor is the Nth feature of Major. If you really want a triple or other tuple of category counters, you are free to create a spec and reference implementation.

Where in the spec does it say that the patch has to be the Nth version of Major.Minor and minor has to be the Nth feature of Major? Consider that the authors can skip version numbers, those interpretations are too strict to be practical.

See, for example, Windows NT kernel version bump from 6.3 (marketed as Windows 8.1) to 10.0.

jwdonahue:
@Explorer09 it's the effective implication of the reset rules.

As currently formulated any version string that adheres to this spec, tells you how many breaking release are behind that version, how many features have been released since the last breaking change, and how many patches have been released, since the last feature release.

I would argue that you should have an alternative spec, that states that version fields MUST NOT be reset to zero, when a more significant field is bumped. If it is either/or, then we really don't know which semantics that string represents.

Most, if not all specs, have unstated implications.

jwdonahue:
@Explorer09

You likely won't win many arguments here, pointing at non-conforming versioning practices. There are many more instances of non-conforming practices in the real world, than there are SemVer conforming instances.

NT Kernel, Windows and many of Microsoft's products, do not adhere to SemVer. In some cases, there are no externally consistent semantics for their products, other than the build number. Internally, they have their own set of practices, including SemVer. 

I was among many voices in Microsoft, pushing to adopt SemVer, as far back as 2012. My team (Windows Build at the time), was instrumental in pushing the Nuget team to adopt SemVer. They dragged their feet initially, so we invented a new SemVer compliant tool, to package our internal build tools and other binary assets, as we moved the entire codebase to git.

Explorer09:
> @Explorer09 it's the effective implication of the reset rules.
> 
> As currently formulated any version string that adheres to this spec, tells you how many breaking release are behind that version, how many features have been released since the last breaking change, and how many patches have been released, since the last feature release.

This assumption has been removed in this commit:
https://github.com/semver/semver/commit/2bfa9b0cbac8dcc33fc836ab4e5162bf014e19ff

Originally it does require that the increment has to be done by one. It no longer is.

> You likely won't win many arguments here, pointing at non-conforming versioning practices. There are many more instances of non-conforming practices in the real world, than there are SemVer conforming instances.

I'm not trying to "win" an argument here. It's more of a criticism of a spec - that this spec is too strict. Of course I could fork the spec and create my own scheme for my own use. But then, how would it be beneficial to other would-be adopters of Semantic Versioning? That's the purpose of a community.

jwdonahue:
@Explorer09

> This assumption has been removed in this commit:
https://github.com/semver/semver/commit/2bfa9b0cbac8dcc33fc836ab4e5162bf014e19ff

Yes, you're right, it's not necessarily an exact count of actual patches or features, but they are always related to the more significant field in the string. There's a lot of automation out there that depends on those relationships.

I don't recall why that loophole was added at the time. I do know there are a lot of CI/CD automation systems out there, that always increment by 1, but don't necessarily release every version to the public. If you had a package service that refused to accept 1.0.2, before you published 1.0.1, it would be problematic, especially for systems that publish to multiple package feeds, for more reasons than I can enumerate.

It is unlikely that anyone is generating 1.0.999 immediately after 1.0.0. Most of us do take the rate of features and patches per unit time, to be an indicator of stability. If the publisher has policy of always publishing at higher increments, we can generally take that into account.

The loophole does alleviate some the pressure from proponents of adding a 4th field, so that they can patch their patches. I think I've seen some advice handed out there that suggested skipping 5, 10 or 100, so you could go back patch a patch, but I think if you're system is that complicated, you might as well drop SemVer in favor of one of the many other perfectly useable versioning schemes.

jwdonahue:
@Explorer09 

> I'm not trying to "win" an argument here. It's more of a criticism of a spec - that this spec is too strict.

Apologies, my assumption was based on the fact that your comments seem to be related to the OP's request to make a change to the spec.  It seems to me, to be pointless to criticize the spec, or the proposed changes, if you are not in fact arguing one way or the other.

It's essentially, the whole point of these threads.

jwdonahue:
Here's a simple history to consider:

...
1.0.1 // Released.
...
2.0.0 // Released.
2.0.1 // Released fixed pixel.
1.0.2 // Released fixed pixel.
...
3.0.0 // Released.
1.0.3 // Released, fixed pixel for real this time.
// No additional pixel fix required for 2.y.z
3.1.0 // Add jump through hoops feature.

How does that work with the OP's proposed change?

Consider:

1.2.1
2.3.1
1.?.2 // Bug fix on v1 branch, what should I do with the minor field?

I don't know, maybe there's a rule that we can add to the spec, that allows the reset feature to be optional? What impact will that have on existing automation? It's easy enough to wave your hand and declare "see, I made this work manually", but what about the thousands of implementations in the real world, that might suddenly be presented with these changes?

What about parallel branch and/or build/release systems?

What about all those systems that store the version string in the repo, and pull whatever tag was attached to the commit that the changes were applied to, to determine what the next tag should be?

I think I can make the OP's scheme work, if I maintain a separate database of product version strings, mapped to their associated source content hashes, but that's a "best practice" process that the vast majority of CI/CD systems do not support.

jwdonahue:
I would also add that there are many consumers out there, who never take X.0.0, until it's aged a bit. If you don't force the reset, it's a little harder to tell whether the current major version has had it's first set of bug fixes.  Given the history of OS's back in the 90's and early 2000's, I wouldn't be surprised if that one thought wasn't the main driver behind the reset requirement. But I think it also tends to keep the version strings a little bit shorter over the life of a long lived product.

jwdonahue:
@sgryphon 

> Going through my old PRs and found that although this had been approved it never merged.

It has not been approved. See https://github.com/semver/semver/blob/master/CONTRIBUTING.md

Explorer09:
> Here's a simple history to consider:
> 

> How does that work with the OP's proposed change?
> 
> Consider:
> 
> 1.2.1
> 2.3.1
> 1.?.2 // Bug fix on v1 branch, what should I do with the minor field?
> 

Not sure who you are talking with, but I think the answer to this one is straightforward: 1.2.2.

Alternatively 1.3.2 can work as well if a non-breaking feature is added to v1 branch.

It's better to not overcomplicate the original idea of SemVer, namely that a major version bump always mean a breaking change. There's no requirement that a developer cannot release a new 1.x version after a 2.x.y is released.

The real requirement is that, after a developer releases a 2.x, any subsequent 1.x releases MUST NOT introduce a breaking change. If there has to be an API breakage in 1.x, just kill or EOL (end-of-life) the 1.x branch.

jwdonahue:
I may have been too focused on the idea that this change was proposed for the purpose of turning the fields into category counters, which apparently isn't the case here, but one that I would oppose.

I think maybe we have a scope problem here. The spec isn't particularly strong wrt publisher vs package store vs consumer behaviors. It's focused mostly on the syntax, with a little bit of publisher restrictions and sorting guidelines for the rest.

I haven't spent enough time drilling down on this to be sure, but I suspect the proposed change, would require some additional explicitly scoped behavioral specs. No idea whether they would necessarily lead to breaking changes in total, but I do think the OP's original concern was addressed without changing the spec.

I think the reset requirement on the version generator side of things is perfect. We should not carry previous patch value, into the next feature bump, nor should we carry previous feature value, into the next major bump. That keeps the strings as short as possible over time, and it's exactly how most of the tooling works today. 

What keeps coming up in these issues, has more to do with what package services should accept and what consumers might find on those services. I don't much like the idea that the spec should constrain those innovators, except to the extent that they should accept valid SemVer syntax, and ensure that no SemVer string, references more than one package content hash. The current spec adds sorting rules, that I personally think go a little too far, in constraining how lists should be sorted, but I've found it to be a workable solution, except for when I worked around it ;).

Explorer09:
> 
> I think the reset requirement on the version generator side of things is perfect. We should not carry previous patch value, into the next feature bump, nor should we carry previous feature value, into the next major bump. That keeps the strings as short as possible over time, and it's exactly how most of the tooling works today.

Nope. AFAICT, SemVer gives no guarantee that the version numbers would be short. It is possible for a major version numbers to be as high as 100 or 200. The requirements of making version strings short is out of scope of this spec, and it's good to keep it that way.

sgryphon:
Hi. The proposed change is simply to allow, where necessary for ".0" versions to be skipped. (The same as you can already skip ".n" versions if you want; why can you skip ".n" versions but not ".0").

This doesn't break any of use of the spec or the purpose of the spec. Version 2.1.2 is still greater that 1.8.3, and at that point whether or no 2.0.0 ever existed or not should be irrelevant.

----

The example posted on Dec 7, 2017 by jwdonahue (asking me what the problem is), illustrates the issue perfectly.

The spec says (in rule 7) "Patch version MUST be reset to 0 when minor version is incremented."

The example questioning why I wanted to change that (to SHOULD) included these two lines:

3.0.2-beta // Ships.
3.1.3-beta // Ships, marketing insisted on one late feature.

Technically the example does NOT conform to SemVer, because when the minor version incremented from 0 to 1, the patch was NOT reset to 0.

I think this is a silly requirement, and the example provided was perfectly fine.

----

Another example, from Explorer09 earlier today.

Given the example history:
1.0.1 // Released.
2.0.0 // Released.
2.0.1 // Released fixed pixel.
1.0.2 // Released fixed pixel.
3.0.0 // Released.
1.0.3 // Released, fixed pixel for real this time.
3.1.0 // Add jump through hoops feature.
1.2.1
2.3.1

According to the current spec 1.2.1 is non-conforming, because there is no 1.2.0

Similarly 2.3.1 is non-conforming, because there is no 2.3.0

Explorer09's statement "Alternatively 1.3.2 can work as well if a non-breaking feature is added to v1 branch."|

Is incorrect, i.e. does not conform to the current spec. 1.3.2 is not allowed because there is no 1.3.0.

When the minor version increases (there are some 1.2 versions, so increase to 1.3), the patch MUST (current spec) reset to 0.

You are welcome to skip 1.3.1, because the spec says they must increase, so 1.2.1 -> 1.3.0 -> 1.3.2 is perfectly fine. (but 1.2.1 -> 1.3.1 is not)

----

There are problems with "hey our build server created a 1.3.0, but we never published it", or "we pulled 1.3.0 from the release", because an external system that is validating and exactly following the spec should deem you invalid.

In practice there is no way for it to know whether 1.3.0 really exists or not. Maybe you created a dummy version purely existing in your head. Trying to enforce "MUST reset to 0" would be counter productive, and probably not what users of the spec exist.

Is there anyone here seriously arguing that going from version 3.0.2-beta to 3.1.3-beta, or from version 1.2.1 to 1.3.2, should be disallowed and make your application non-conforming? (Because right now it technically is)

Changing the spec to say "Patch version SHOULD be reset to 0 when minor version is incremented.", makes it clear that it is the normal and expected practice, but if for some reason you don't, then no shenanigans are needed.

ljharb:
I'm still unclear as to why this change would even make sense. If you never published 1.3.0, *it doesn't exist*, and you can still publish it. Nonpublished versions aren't immutable and shouldn't be preserved anyways.

> should be disallowed and make your application non-conforming

yes. don't do that.

Explorer09:
> Explorer09's statement "Alternatively 1.3.2 can work as well if a non-breaking feature is added to v1 branch."|
> 
> Is incorrect, i.e. does not conform to the current spec. 1.3.2 is not allowed because there is no 1.3.0.

Clarification. I only said "working", not "conforming". In fact I agree that the requirement of resetting the patch version to 0 is silly.

SemVer should only mandate the ordering of the version numbers, not whether there must be a x.y.0 whenever an x.y.1 version exists.

jwdonahue:
The reset rule applies to version generators, NOT to the publicly available history, on a package service. Publishers and feed services, MUST be allowed to remove versions from their feeds. There are technical and legal reasons for this. That does not imply that tooling should be allowed to ignore the reset rule.

...
1.1.1 // Current version.
// Add a feature
1.2.0 // Published, NOT 1.2.1
// Somebody notices you've ripped LGPL code and rereleased it as MIT or some other non-LGPL compatible license.
1.2.0 // Removed from public view.
// Fix the issue.
1.2.1 // Published here, we cannot reuse 1.2.0

Currently available history on any feeds for this product:
...
1.1.1
1.2.1

Once you've published 1.2.0, you can never apply that version to any subsequent release/publication. If you use version tags in your repo, which was a common practice at the time of the original spec, you're not allowed to move that tag to another commit Id. You cannot have a P 1.0.0 with hash N, and a P 1.0.0 with hash M. There be dragons if you go there!

I often work with manual and CI/CD systems that generate Nuget packages on every single build. Each of those packages MUST have a version attached to them. Once created, they tend to be "published" for various other internal processes to consume, before any of them are ever released to a public feed. 

What some consumers see, might include a good many gaps in the version sequence. There's no reason for the spec to forbid that practice and I don't think the spec addresses how feeds are managed anyway, so no problem with the above practice. We adhere strictly to the SemVer rules when generating package versions, and we rely on them having been followed, when we decide what versions to consume.

What the spec implies, is that a version generating tool, must account for the highest version it's ever generated for a specific product, and follow the increment rules, when asked to generate the next one. If you think that is silly, then, go ahead and ignore the spec. You are not required to be SemVer compliant, unless you actually claim to be SemVer compliant.

jwdonahue:
@Explorer09 
> SemVer should only mandate the ordering of the version numbers, not whether there must be a x.y.0 whenever an x.y.1 version exists.

The spec does not say that at all. It says that if you have an x.y.z, the next patch number you generate must be x.y.z+1, or the next minor number to generate must be x.y+1.0.

------------------

Correction: ...next patch number you generate must be x.y.z+N, or the next minor number to generate must be x.y+N.0, where N > 0.

jwdonahue:
@Explorer09 

> Nope. AFAICT, SemVer gives no guarantee that the version numbers would be short. It is possible for a major version numbers to be as high as 100 or 200. The requirements of making version strings short is out of scope of this spec, and it's good to keep it that way.

I never said the there was any such requirement. I merely pointed out that the practice of resetting the right hand fields to zero, tends to keep the strings shorter. I have seen 4 and 6 digit numbers used in some fields. I did a survey on that a number of years back, and the vast majority of the fields were single digit.

I agree, there is no version string length limit in the spec, nor should there be.

jwdonahue:
I think the intent of the reset rule, is to ensure that patch is not carried into the next minor release and minor is not carried into the next major release. So 1.0.99, does not lead directly to 1.1.99. In other words, the patch version is related to the minor version, is related to major. They are not independent event counters.

Can we agree on that?

Explorer09:
> @Explorer09
> 
> > SemVer should only mandate the ordering of the version numbers, not whether there must be a x.y.0 whenever an x.y.1 version exists.
> 
> The spec does not say that at all. It says that if you have an x.y.z, the next patch number you generate must be x.y.z+1, or the next minor number to generate must be x.y+1.0.
> 
> Correction: ...next patch number you generate must be x.y.z+N, or the next minor number to generate must be x.y+N.0, where N > 0.

This thing deserves a special clarification in the spec. It's a discrepancy between the producer of a version number and the consumer of it. The requirement for a producer can be strict, but consumer should be tolerant on what version numbers it can accept.

For the producer side you can say "when the minor number is incremented, automated version generators SHOULD reset the patch number to 0" (but note that producers failing to do so is not a strict violation of the spec). For the consumers, however, there should be an indication that they MUST accept skipping version numbers and not treat those as errors.

The current wording of the spec didn't suggest tolerance on the consumer side.

Explorer09:
> I think the intent of the reset rule, is to ensure that patch is not carried into the next minor release and minor is not carried into the next major release. So 1.0.99, does not lead directly to 1.1.99. In other words, the patch version is related to the minor version, is related to major. They are not independent event counters.
> 
> Can we agree on that?

No. That assumption is not the business of SemVer. When I maintain software, it is my decision on whether the patch version will reset or will be monotonically increasing. Customer should only tell that 1.0.99 < 1.1.99, not whether there would be a 1.1.98 or whatever. If that makes the patch version very large, so be it.

I know that in certain situations having a large version number could be a bad idea. E.g. [library soname versions](https://man7.org/conf/lca2006/shared_libraries/slide5d.html). But AFAICT, no linkers for now would reject linking with libraries with large minor or patch numbers yet.

ljharb:
Of course it's the business of semver, as is anything in the spec. If you don't like it, you don't have to comply with it.

If your users can't understand that `1.1.0` is larger than `1.0.9999999` then they won't care about semver compliance, either.

jwdonahue:
> No. That assumption is not the business of SemVer. When I maintain software, it is my decision on whether the patch version will reset or will be monotonically increasing. Customer should only tell that 1.0.99 < 1.1.99, not whether there would be a 1.1.98 or whatever.

In that case, you are not SemVer compliant, and that's okay. You don't have to be SemVer compliant. The fact that you are not, and that it probably doesn't break anybody who uses standard tooling, is not sufficient justification to change the spec.  Changing it to give you a warm fuzzy about your uncommon practices, will not make it any less annoying to your consumers.

The spec was intended to document one of, what had become a common practice, at least as far back as the 90's. I don't know if @mojombo ever looks in on these threads, these days, but I'd be willing to bet a donut and a cup of coffee, that he intended to steer developers away from that exact process.

jwdonahue:
> ...The requirement for a producer can be strict, but consumer should be tolerant on what version numbers it can accept.

That is a common protocol design pattern. Consumers must expect all manner of failures. In this case, the consumers are interacting with one or more package/file store/feed, that is often remote and potentially transient. The version generator can be spec'd to more or less high precision, such that it's outputs are far more predictable, than what you might find on the internet or a LAN.

-----

Consider how consumers usually interact with the version set. It's probably pretty rare that they would attempt to pull an explicit version. Generally they use whatever range selection mechanism, their particular packaging silo implements. In the off case that they do expect a specific x.y.z, they must be prepared to fail for any number of reasons, including the prospect that x.y.z is no long publicly available, or possibly, never was. The SemVer spec doesn't need to say that, because anyone who ever built any non-trivial system of any kind, knows in their bones, they must handle failures gracefully.

jwdonahue:
And I would add that, the publish step, comes after that package generation step, where the version number is often required. The spec does not and should not attempt to regulate our entire build/release processes. There's just too many of them out there and many of us have multiple levels of "publication".

The set of versions available for P, are not always the same across all feeds. It would be impractical to require that.

Explorer09:
> > No. That assumption is not the business of SemVer. When I maintain software, it is my decision on whether the patch version will reset or will be monotonically increasing. Customer should only tell that 1.0.99 < 1.1.99, not whether there would be a 1.1.98 or whatever.
> 
> In that case, you are not SemVer compliant, and that's okay. You don't have to be SemVer compliant. The fact that you are not, and that it probably doesn't break anybody who uses standard tooling, is not sufficient justification to change the spec. Changing it to give you a warm fuzzy about your uncommon practices, will not make it any less annoying to your consumers.

This argument is a waste of my time. Quote the reply I made just before:

> I'm not trying to "win" an argument here. It's more of a criticism of a spec - that this spec is too strict. Of course I could fork the spec and create my own scheme for my own use. But then, how would it be beneficial to other would-be adopters of Semantic Versioning? That's the purpose of a community.

jwdonahue:
@Explorer09, Okay:

> I'm not trying to "win" an argument here. It's more of a criticism of a spec - that this spec is too strict. Of course I could fork the spec and create my own scheme for my own use. But then, how would it be beneficial to other would-be adopters of Semantic Versioning? That's the purpose of a community.

Criticism, in this forum is tantamount to arguing for or against whatever you're criticizing.  I don't think you or the OP have shown the need for the proposed change.

Like many other alternative versioning schemes, your favorite flavor, probably has adherents. I know of a handful of version/revision specs and best practices, that look nothing like SemVer, though some do have semantics of a sort. I have often used them where SemVer would not be appropriate or necessary. I would not criticize something, simply because it does not meet a particular need, unless I was being forced to use that something and said criticism was justified.

The reset rule is one of the features of SemVer that sets it apart from many other schemes.

> This argument is a waste of my time.

Yes.

Explorer09:
> […] unless I was being forced to use that something and said criticism was justified.

This came out of your mouth. You admitted there is a force.

> 
> The reset rule is one of the features of SemVer that sets it apart from many other schemes.

But then the OP's question remains: Why the keyword MUST? Why alienate edge cases or special situations?

jwdonahue:
> But then the OP's question remains: Why the keyword MUST? Why alienate edge cases or special situations?

I already answered that question to the best of my ability. I know why I want it to remain as is, I can only surmise why the original author of the spec, made it so. Like I said, I think the reset rules are there to set SemVer apart from the many 'category of counters', versioning schemes out there.

>> […] unless I was being forced to use that something and said criticism was justified.

> This came out of your mouth. You admitted there is a force.

Now I think you're just being argumentative for the sake of it.

PR #399, opened 2017-10-12
Proposal for 2.1: Explicitly declare forward- and backward-compatibility

brandtbucher:
## Taken from the proposed changes:

Don't be fooled by the new version number; while backward-compatible (of course), version `2.1` fundamentally changes the way SemVer works. It is recommended that you read this documentation thoroughly.
 
In short, `2.1` lets you safely use more versions of your dependancies than `2.0.0` would, without breaking any code. Its implementation is the purest form of knowing exactly which versions can act as complete substitutes for any other version. `2.0.0` doesn't quite allow for this, because its hard cap at 3 places of version precision inherently loses information about forward and backward compatibility as more and more minor-level changes are added. As you will see, `SemVer 2.1` improves upon this model by explicitly declaring forward- and backward-compatibility with any other version in an easy-to-understand way.

Given a hypothetical version `1.2.3`:

- If making changes that are both forward- and backward-compatible with the current version (i.e. a bug fix), add `.1` to the far right of the version number. In this case: `1.2.3.1`.
- If making changes that are backward-compatible but not forward-compatible with the current version (i.e. an addition to the API), increment the version number. In this case: `1.2.4`.
- If making changes that are not backward-compatible with the current version (i.e. an API removal), increment the shortest part of the version number for which the changes are not backwards compatible, and drop the rest.
  - In this case, assuming the change is backwards-compatible with version `1` (but not version `1.2`): `1.3`.
  - Alternately, assuming the change isn't backward-compatible with any prior version: `2`.
  
So, for example, if you used you used version `1.2` of our hypothetical piece of software in your code, you know that any version beginning with "1.2" will function fully as a drop-in replacement. So, in the above examples, `1.2.3`, `1.2.3.1` and `1.2.4` will work, but `1.3` and `2` won't. However, the true
beauty of `SemVer 2.1` lies in its granularity: because version numbers can be of arbitrary length, you are able to know exactly which versions are compatible with (theoretically infinite) precision.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -287,7 +287,7 @@ impact of your changes, and evaluate the cost/benefit ratio involved.
 
 You might, but you don't necessarily have to. However, in a world of insanely
 long cryptographic keys and million-line programs, it is not unreasonable to see
-the inherent usefulness of infinitely precice versioning. SemVer has never
+the inherent usefulness of infinitely precise versioning. SemVer has never
 claimed that its goal is to create "short" version numbers. Rather, it aims to
 create useful, consistent version numbers. `SemVer 2.1` is just the latest step
 in the realization of this goal. However, you are allowed to reduce this
@@ -296,7 +296,7 @@ number at some arbitrary point, and drop the rest. So, the above version could
 be shortened to `2`, or `1.2`, or `1.1.2`, or `1.1.1.2`, etc... Just know that
 as the version number becomes shorter and shorter, fewer and fewer projects will
 deem the new version "safe" to use. Even if the version number is kept at 3
-places (as in `SemVer 2.0.0`), `SemVer 2.1` still commuinicates the same (or in
+places (as in `SemVer 2.0.0`), `SemVer 2.1` still communicates the same (or in
 many cases, more) information about compatibility than earlier numbering schemes
 did.
 


brandtbucher:
I’d like to add one more note on why I chose to call the new version `2.1` instead of `3`, which is what it should be called (according to its own specs). Honestly, I went back and forth on this quite a bit. However:

1. I determined that `2.1` would most effectively communicate the backward-compatibility for the new version to both groups (those using `2.0.0.` and those using the new version).
1. I also feel as if SemVer, against its own rules, has no publicly declared API. If we determine that SemVer’s API is defined as the specs it outlines, then fine, it should be `3`. But if we define its API as what SemVer actually *does*, this really doesn’t change that functionality—it just makes it better. So, as I see it, it’s undefined what constitutes a forward-incompatible API change for SemVer.
1. Finally, because it introduces the new version, the inconsistency of the name `2.1` would not cause problems for any existing `2.1` users... There are none!

The ultimate purpose of SemVer is communication of changes and compatibility. With this in mind, I went with `2.1`.

jwdonahue:
@brandtbucher 

You are proposing breaking changes to the spec. Any scheme that adds to or removes fields from the version triplet, will break existing implementations.  The current vibe I am getting from a read of the issues here, is that another major version bump would be destabilizing. There are package managers out there that barely cover version 1.0.0 correctly and we already have version 2.0.0 and there is still room for improvement without breaking the world. All previous attempts to start a 3.0.0 branch, that I am aware of, have been rebuffed.  I do not anticipate that your proposed changes will ever be accepted.

FichteFoll:
Could you explain why forwards compatibility conveys any more information than "not increasing major" does?

brandtbucher:
@jwdonahue 

I totally understand that the chances of my branch being merged are very low, given SemVer's historical resistance to change. I guess you could call it more of an *idea* I had, which I proposed as a pull request, because, well, why not? 3 places of versioning precision just seems arbitrary to me, and it took less than an hour to begin forming an interesting alternative (or modification, depending on how you look at it). Besides, I'd say that, in general, low adoption rate of *any* standard would mean a somewhat painless (and, perhaps even *necessary*) transition.

Of course, this proposal is obviously not perfect or really even complete. But like I said - why not?

---

@FichteFoll 

Of course.

To recap, because the version numbers (under this proposal) can extend to any length, the choice of *which* part of the number is incremented or extended can mean different levels of forward- or backward-compatibility based on what exactly you are comparing. Say I have some dependency:

**Backward-compatibility** means that  I can safely upgrade to the newer version without breaking anything. In other words, **no functionality was *removed* or *changed***.

**Forward-compatibility** is a term you tend to see less often. It means that  I can safely downgrade to the older version without breaking anything. In other words, **no functionality was *added* or *changed***.

Under the current SemVer specs, **you cannot tell with any granularity from the version number alone whether or not you can safely accept older versions of dependencies**. Without testing every minor version individually, the basic thinking becomes:

- I shouldn't allow anything with a lower patch number, because it has bugs.
- I have no idea if I can allow anything with a lower minor version, because It may remove functionality that I need.
- And, duh, I obviously can't allow anything with a different major version!

Under the new proposal, you can tell exactly which past and future versions of a dependency can safely be used *from the version number alone*.

Just one example:

Imagine that I am using a very outdated version of some dependency. Many new versions have come out without me knowing, each adding new functionality and new bug fixes. However, the newest version introduced a **huge** breaking change to some new functionality that my older version never even had.

With SemVer, the incremented major version number would scare me away from this newer version, because I would assume that I need to significantly rework my code first. As a result of this, I could be missing out on great new functionality, or even critical security fixes!

Under the new proposal, one quick glance (or automated scan) would instantly make it clear that it could function fully as a drop-in replacement for my outdated version, because of the *granularity* with which it allows declarations of backwards compatibility.

brandtbucher:
And the best part: if it gets accepted and you don't like it, you can just keep using SemVer as before!

Sure, you won't get any of the newer benefits by following the guidelines above, but 2.0.0-style numbering still follows the new relaxed rules just fine, since they allow manual overrides for the purposes of number shortening, communication, etc.

jwdonahue:
@brandtbucher, please make up a "version log" of release events that show your versioning scheme in use. In other words something like:

0.0.1    // Experiment shared with my team.
0.1.2    // Converging on prototype for the entire org.
1.0.0    // First release

Etc., only use your version notations and conjure up scenarios that cover its feature set.  Let's just ignore prerelease and build meta tags for the time being.

brandtbucher:
@jwdonahue 

Great idea! Here’s a lifetime log of a package I just made up, with comparisons of the current and proposed numbering schemes. I think it makes a many of the potential advantages clear:

| 2.1       | 2.0.0 | Change Log          |
|-----------|-------|---------------------|
| 0.1       | 0.1.0 | Initial development |
| 0.2       | 0.2.0 | New feature A       |
| 0.3       | 0.3.0 | New feature B       |
| 0.3.1     | 0.3.1 | Bug fix             |
| 0.3.2     | 0.4.0 | New feature C       |
| 0.3.3     | 0.5.0 | New feature D       |
| 0.3.3.1   | 0.5.1 | Bug fix             |
| 0.3.3.1.1 | 0.5.2 | Bug fix             |
| 1         | 1.0.0 | Stable release      |
| 1.1       | 1.0.1 | Bug fix             |
| 1.1.1     | 1.0.2 | Bug fix             |
| 1.1.2     | 1.1.0 | New feature E       |
| 1.1.2.1   | 1.1.3 | Bug fix             |
| 1.1.2.2   | 1.2.0 | New feature F       |
| 1.1.2.2.1 | 1.3.0 | Deprecate feature E |
| 1.1.3     | 2.0.0 | Remove feature E    |
| 1.1.4     | 2.1.0 | New feature G       |
| 2         | 3.0.0 | Break feature C     |
| 2.1       | 3.0.1 | Bug fix             |
| 2.1.1     | 3.0.2 | Bug fix             |
| 2.1.1.1   | 3.0.3 | Bug fix             |
| 2.1.1.2   | 3.1.0 | New feature H       |
| 2.1.1.3   | 3.2.0 | New feature I       |
| 2.1.1.4   | 4.0.0 | Break feature H     |
| 2.1.1.5   | 5.0.0 | Remove feature I    |
| 2.1.1.6   | 6.0.0 | Remove feature H    |
| 2.1.1.6.1 | 6.1.0 | End-of-life warning |
| 3         | 7.0.0 | Nonfunctional shell |

brandtbucher:
Basically, an important thing to remember is that a breaking change does not necessarily affect all past versions. And the current SemVer implementation ignores that.

jwdonahue:
@brandtbucher, perfect! Now we have some points of reference.  So out of that history, which earlier versions are forward compatible with which future versions?

jwdonahue:
@brandtbucher, sorry, I just can't let you do all that work before you give some serious thought to the following question:

What if I add 35 features one at a time while permanently maintaining back-compat?  What if it's 100?  How many dotted fields do I wind up with?  Please define the state machine I have to implement to tell me whether I can safely roll-back from 2.1.1.5 (5.0) to 2.1.1.3 (3.2.0), without breaking my project that depends on your module.

If you can complete those tasks, I'll take a serious look at your scheme.  My current view is that cognitive load on human consumers of your scheme is too high.  Please prove me wrong.

brandtbucher:
@jwdonahue 

> So out of that history, which earlier versions are forward compatible with which later versions?

Easy. If you have 1s on the right side of your version number, chop them off one-by-one (ha, get it?) to get all compatible versions. Here:

| 2.1       | 2.0.0 | Safe to downgrade to: |
|-----------|-------|---------------------|
| 0.1       | 0.1.0 |  |
| 0.2       | 0.2.0 |  | 
| 0.3       | 0.3.0 |  |
| 0.3.1     | 0.3.1 | 0.3 |
| 0.3.2     | 0.4.0 |  |
| 0.3.3     | 0.5.0 |  |
| 0.3.3.1   | 0.5.1 | 0.3.3 |
| 0.3.3.1.1 | 0.5.2 | 0.3.3, 0.3.3.1 |
| 1         | 1.0.0 | |
| 1.1       | 1.0.1 | 1 |
| 1.1.1     | 1.0.2 | 1, 1.1 |
| 1.1.2     | 1.1.0 | |
| 1.1.2.1   | 1.1.1 | 1.1.2 |
| 1.1.2.2   | 1.2.0 |  |
| 1.1.2.2.1 | 1.3.0 | `1.1.2.2` |
| 1.2       | 2.0.0 |  |
| 1.3       | 2.1.0 |  |
| 2         | 3.0.0 |  |
| 2.1       | 3.0.1 | 2 |
| 2.1.1     | 3.0.2 | 2, 2.1 |
| 2.1.1.1   | 3.0.3 | 2, 2.1, 2.1.1 |
| 2.1.1.2   | 3.1.0 |  |
| 2.1.1.3   | 3.2.0 |  |
| 2.1.2   | 4.0.0 |  |
| 2.2   | 5.0.0 |   |
| 3   | 6.0.0 |  |
| 3.1 | 6.1.0 | `3` |
| 4         | 7.0.0 |  |

(I found them for all 28 versions in one minute.)

> You might as well also list the later versions that are backwards compatible with the earlier versions as well.

No problem. Drop the right-hand number from your current version (unless the whole version is only one part). Anything that starts with what you have left over is fair game. Again:

| 2.1       | 2.0.0 | Safe to upgrade to: |
|-----------|-------|---------------------|
| 0.1       | 0.1.0 | 0.2, 0.3, 0.3.1, 0.3.2, 0.3.3, 0.3.3.1, 0.3.3.1.1 |
| 0.2       | 0.2.0 | 0.3, 0.3.1, 0.3.2, 0.3.3, 0.3.3.1, 0.3.3.1.1 | 
| 0.3       | 0.3.0 | 0.3.1, 0.3.2, 0.3.3, 0.3.3.1, 0.3.3.1.1 |
| 0.3.1     | 0.3.1 | 0.3.2, 0.3.3, 0.3.3.1, 0.3.3.1.1 |
| 0.3.2     | 0.4.0 | 0.3.3, 0.3.3.1, 0.3.3.1.1 |
| 0.3.3     | 0.5.0 | 0.3.3.1, 0.3.3.1.1 |
| 0.3.3.1   | 0.5.1 | 0.3.3.1.1 |
| 0.3.3.1.1 | 0.5.2 |  |
| 1         | 1.0.0 | 1, 1.1, 1.1.1, 1.1.2, 1.1.2.1, 1.1.2.2, 1.1.2.2.1, `1.2`, `1.3` |
| 1.1       | 1.0.1 | 1.1.1, 1.1.2, 1.1.2.1, 1.1.2.2, 1.1.2.2.1, `1.2`, `1.3`  |
| 1.1.1     | 1.0.2 | 1.1.2, 1.1.2.1, 1.1.2.2, 1.1.2.2.1 |
| 1.1.2     | 1.1.0 | 1.1.2.1, 1.1.2.2, 1.1.2.2.1 |
| 1.1.2.1   | 1.1.1 | 1.1.2.2, 1.1.2.2.1  |
| 1.1.2.2   | 1.2.0 | 1.1.2.2.1 |
| 1.1.2.2.1 | 1.3.0 |  |
| 1.2     | 2.0.0 | 1.3 |
| 1.3    | 2.1.0 |  |
| 2         | 3.0.0 | 2.1, 2.1.1, 2.1.1.1, 2.1.1.2, 2.1.1.3, `2.1.2`, `2.2` |
| 2.1       | 3.0.1 | 2.1.1, 2.1.1.1, 2.1.1.2, 2.1.1.3, `2.1.2`, `2.2` |
| 2.1.1     | 3.0.2 | 2.1.1.1, 2.1.1.2, 2.1.1.3, `2.1.2` |
| 2.1.1.1   | 3.0.3 | 2.1.1.2, 2.1.1.3, |
| 2.1.1.2   | 3.1.0 | 2.1.1.3, |
| 2.1.1.3   | 3.2.0 | |
| 2.1.2   | 4.0.0 | |
| 2.2   | 5.0.0 | |
| 3   | 6.0.0 | `3.1` |
| 3.1 | 6.1.0 |  |
| 4         | 7.0.0 |  |

(It took me less than four minutes to do them all.)

> What if I add 35 features one at a time while permanently maintaining back-compat? What if it's 100? How many dotted fields do I wind up with?

I think you meant to ask a harder question than you did. Adding functionality (forward-incompatible, backward-compatible changes) doesn't extend the number. It increments it. So if I started with `1.2.3`, then after 35 new features (added one at a time), it would be `1.2.38`. After 100, it would be `1.2.103`. **The number of dotted fields is exactly equal to the number you started with.**

However, I'll do my extra credit and assume you mean forward- and backward-compatible bug fixes. Yes, you would end up with a version 35 or 100 places longer than originally, filled with all 1's. However, the specs allow you to truncate and increment wherever you want (this is also how we keep compatibility with SemVer 2.0.0). So, after 99 bug fixes, for the next version number I could just truncate to, say, three places, and increment that 1 to a 2. It would contain *exactly as much information as a SemVer version number*. If I was feeling wild and crazy and truncated it at, say, *four* places of precision, I am now communicating **more** information about compatibility with other versions than the old SemVer ever could! 

> Please define the state machine I have to implement to tell me whether I can safely roll-back from 2.1.1.5 (5.0) to 2.1.1.3 (3.2.0), without breaking my project that depends on your module.

Writing such a program (to check downgrade compatibility) would be easy - just check my answer to your first question! However, I'll formalize the algorithm here:

- Remove all trailing 1s from both numbers.
- If they are the same, you can downgrade. If not, you can't.

> If you can complete those tasks, I'll take a serious look at your scheme. My current view is that cognitive load on human consumers of your scheme is too high. Please prove me wrong.

I left the SemVer 2.0.0 numbers in the tables above in order to illustrate something. It is just as easy to answer the "difficult" questions you posed with 2.1 as it is with 2.0.0. And, you get *more* answers (which I `highlighted`, indicating that 2.0.0 would have deemed them incompatible), meaning greater promise of compatibility overall.

brandtbucher:
There are tradeoffs, of course. But the increase in functionality more than makes up for the small learning curve for the new system.

jwdonahue:
@brandtbucher, wow, thanks for the effort!  Are you using this scheme in a real dev-ops environment?

brandtbucher:
Yes, in academic projects mostly.

jwdonahue:
And what was/is the largest number of fields the version got up to?

brandtbucher:
I believe six, with no manual truncating. To estimate it for yourself, just take whatever your highest patch number on a version was, ever. That’s around how many extra decimal places you’ll use.

brandtbucher:
Upper limit, of course.

brandtbucher:
Also, if you’re the sort of person who keeps complete, well formatted change logs, you can automate tools to automatically give you the new version number based on your stated changes. You can even use the change log and version numbers together to automatically find exactly what features and bug fixes you would by losing with any upgrade / downgrade.

jwdonahue:
Ah, well, I've seen patch numbers increment into the tens of thousands.  This is a very interesting scheme though.  I think it's perfect for the purpose of helping machines reason about arbitrary changes in the dependency graph.  

Are you familiar with the diamond point dependency problem?  A => B & C, B =>C,  a bug fix in B forces an update to latest C, the question is, is A compatible with the new C? Well, that's the simplified version of it.  I used to work on a build system where the graph had hundreds of thousand of nodes and a non-trivial number of those were third party libraries and close to half were run-time dependencies!  It's a very interesting problem with far better minds trying to solve it than mine, but they haven't really succeeded yet, that I am aware of.  You wind up tracking content hashes of everything and simply rebuilding anything that depends on an object with a new content hash.  Not very efficient.

You should create a repo and write up the full spec for this idea.  I'd be happy to contribute to the effort. I don't think it's an idea that should be taken on by SemVer just yet, but maybe if it gets some meat on its bones and a few non-trivial users?  Tooling is critical, even if it's just a handful of scripts, anything that makes it easy for early adopters to try out.

Send me an email: joseph@odonahue.com, include a link to this thread so I can get oriented. I'll be travelling away from computers for a few days, but when I get back, I'd like to talk.

jwdonahue:
Oops, just realized this is PR thread.  I see your repo now.

FichteFoll:
Thanks for the huge list of examples. This helps immensely with trying to understand the issue.

The gist of what I'm getting from this is that:

1. `.1` versions are special. 
2. New features always increase the last number. That means if there was a bugfix release after the latest feature addition, you end up with more numbers compared to adding features consecutively. I'm unsure what the meaning of this is as opposed to increasing the "feature count" number in comparison to what SemVer does now. 
3. Due to the hierarchical approach you might end up with being able to deprecate feature F without affecting the state of feature E (i.e. one can assume that `1.1.3` is still backwards compatible wrt E) but you can't do vice versa because of the order you made the releases in. This strikes me as a seemingly arbitrary limitation that might put the entire schema to waste if you're unlucky, for severely increased complexity.
3. I don't get your argument regarding your schema being better for specifying forwards compatibility. You say that bugfix releases in current SemVer tell you nothing about forwards compatibility, but I don't see how appending `.1` would do so in your schema.

Basically, the only advantage I see right now is that you don't have to increase the major number as often since you have a more granular "control" over the features you depend on. You end up increasing *some* number in the pipeline, and depending on which that is, a consumer may or may not update easily. I say "control" because it largely depends on the *order* of the features you add, and this leaves a bitter taste imho. Furthermore, a consumer needs to determine the exact number they want to track (more likely to be multiple numbers even) before they can make use of this added granularity and I fear this process could turn out to be rather bothersome.
You might be better off introducing versioning for the features individually.

Finally, as can be seen by the increasing differences between equivalent "2.1" and "2.0.0" versions, I'm strongly opposed to this only being a minor change. This feels like a competing versioning schema rather than an iteration of SemVer itself and thus I don't believe it can be adapted into semver itself, not even in a major update.
(I also see absolutely no reason for choosing this over current SemVer tbh. If this is based on a misunderstanding, please elaborate on it.)

brandtbucher:
Everyone's made some great points here. I'm going to close this since it's become clear that this isn't really viable as a versioning scheme. However, I may move forward with a similar idea as a separate method of automated compatibility tracking. I'll post it here if so.

Thanks all!

PR #402, opened 2017-10-16
Clarifications regarding dots, hyphens, increments and major version …

BenjaminLefoul:
…increment requirements.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -75,9 +75,9 @@ MUST NOT be modified. Any modifications MUST be released as a new version.
 0. Major version zero (0.y.z) is for initial development. Anything MAY change
 at any time. The public API SHOULD NOT be considered stable.
 
-0. Version 1.0.0 defines the public API. The way in which the version number
-is incremented after this release is dependent on this public API and how it
-changes.
+0. Version 1.0.0 SHOULD define the public API. The way in which the version
+number is incremented after this release is dependent on this public API and
+how it changes.
 
 0. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards
 compatible bug fixes are introduced. A bug fix is defined as an internal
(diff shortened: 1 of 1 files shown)


BenjaminLefoul:
For very simple one-file software, this makes it explicitly possible to have all non-trivial (0.0.0 and 1.0.0) versions ever as X.0.0 with X represented as a 4-char string representing any number between 1000 and 9999. This is for better listing presentation.

FichteFoll:
Making 1.0.0 only optionally define an API makes no sense since that is precisely the *purpose* of this version (or any version, in a way). Without a public API, semver has no use.

BenjaminLefoul:
But this does not imply that it is optional to have a public API, quite the contrary as per number 1. It only makes it optional to number said API 1.0.0.

Currently the requirement of (1) followed by the statement in (5) implies that the existence of version 1.0.0 is necessary, which I feel it shouldn't be.

FichteFoll:
What do you gain from not having to number the version that defines the API as 1.0.0? If it's <1, then you are free to do whatever anyway because of (4). If it's >1 then what are you going to compare the previous versions to?

BenjaminLefoul:
I'm only arguing that instead of splitting the version line at point 1 we split it at point X, where X is explicitly defined at the liberty of the developer (with perhaps a default at 1). But yeah, I guess people who don't want version 1.0.0 but still need semver compliance could just have 0.0.z forever and just increment the patch version, but it feels like a hack trying to avoid problems that semver was supposed to solve.

It's not that important.

jwdonahue:
@BenjaminLefoul, if "it's not that important", please cancel/close this PR.  It is highly unlikely to be accepted.

alexandrtovmach:

PR #409, opened 2017-11-19
Fixed punctuation.

jwdonahue:
Fixed instances of periods inside quoted strings that should be outside the quotes.  Also repaired some incomplete sentences.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -18,7 +18,7 @@ Introduction
 ------------
 
 In the world of software management there exists a dreaded place called
-"dependency hell." The bigger your system grows and the more packages you
+"dependency hell". The bigger your system grows and the more packages you
 integrate into your software, the more likely you are to find yourself, one
 day, in this pit of despair.
 
@@ -44,7 +44,7 @@ affecting the API increment the patch version, backwards compatible API
 additions/changes increment the minor version, and backwards incompatible API
 changes increment the major version.
 
-I call this system "Semantic Versioning." Under this scheme, version numbers
+I call this system "Semantic Versioning". Under this scheme, version numbers
 and the way they change convey meaning about the underlying code and what has
 been modified from one version to the next.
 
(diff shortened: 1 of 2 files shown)


Haacked:
Hmm, whether those periods go inside or outside quotations depends on where you live UK vs US: http://www.quickanddirtytips.com/education/grammar/how-to-use-quotation-marks

> The most common question people ask about quotation marks is whether periods and commas go inside or outside, and the answer depends on where your audience lives because in American English we always put periods and commas inside quotation marks, but in British English periods and commas can go inside or outside (kind of like the American rules for question marks and exclamation points).

@jwdonahue are you by chance from the UK? :smile:

jwdonahue:
No, but I find that on average, American English is a misnomer, particularly wrt technical documentation.  Logic dictates that "dependency hell" is a phrase, not a sentence, therefore the quotes should not contain any terminal punctuation.  American English is probably the least spoken/written dialect of English so I prefer to stick to the real thing.

BenLeggiero:
I agree with most of this, except changing that last period to a comma. That creates a run-on sentence.

jwdonahue:
@BenLeggiero, all three clauses are dependent on each other, hardly a [run a sentence](https://en.wikipedia.org/wiki/Sentence_clause_structure#Run-on_sentences).  Generally speaking a run-on sentence is one like this one with multiple clauses and no punctuation.

semver.md
```diff
@@ -206,9 +206,9 @@ to the users of your software. Once these intentions are clear, flexible (but
 not too flexible) dependency specifications can finally be made.
 
 A simple example will demonstrate how Semantic Versioning can make dependency
-hell a thing of the past. Consider a library called "Firetruck." It requires a
-Semantically Versioned package named "Ladder." At the time that Firetruck is
-created, Ladder is at version 3.1.0. Since Firetruck uses some functionality
+hell a thing of the past. Consider a library called "Firetruck", that requires a
```
Haacked: This first comma becomes unnecessary when you restructure the sentence like this because "that requires" modifies "Ladder" and isn't an independent clause. I'd restructure this as:

> Consider a library called "Firetruck" that requires a Semantically Versioned package named "Ladder". At the time that...

I think that flows better.

jwdonahue: Ok, I'll submit another PR shortly.

BenLeggiero: Might I recommend this:
> Consider a library called "Firetruck" that requires a Semantically Versioned package named "Ladder"; at the time that Firetruck is created Ladder is at version 3.1.0.

FichteFoll:
You can just push a new commit to this branch or force-push if you want to get rid of the current commit history instead of creating a new pr.

BenLeggiero:
@jwdonahue "at the time that Firetruck is created Ladder is at version 3.1.0" is not proper grammar for a phrase; it needs a comma after 'created'.

Issue #341, opened 2016-11-01
Transitive dependencies are part of public API and thus should increment version accordingly


Primetalk:
[What should I do if I update my own dependencies without changing the public API?](http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api)

> Use your best judgment. ... it MAY be best to perform a MAJOR version release ...

For Java the best judgment can be more precise.

1. Transitive dependencies are part of public API.

  Let **C** be user's project, **A** be our library and **d** be a transitive dependency. And let's say that we want to use a newer version of **d** - **d'** - in our new version of **A'**. In Java application with the single class path there are two options: either transitive dependency jar be available on the user's classpath in the version **d'**, or it'll be available in the previous version **d** (ceteris paribus). In the former case other modules/libraries that were expecting older version **d** would suffer from incompatibility, in the latter case our library would suffer. The degree of change is exactly equal to the _degree of version change_ of **d** (MAJOR, MINOR, PATCH). Thus **d** is effectively part of the public API of our library **A**.

2. Hence whenever we change a transitive dependency version to some degree, we should change the version of our project to corresponding degree.
  - MAJOR change in **d** => increment MAJOR part of version **A**;
  - MINOR change in **d** => increment MINOR part of version **A**;
  - PATCH change in **d** => increment PATCH part of version **A**.

  Of course, we suppose that **d** follows **semver** recommendations.

Proposal: Amend this FAQ section and add a special notice for Java (as an important domain for semver).

For Java (and similar environments) transitive dependencies are part of the public API and thus the version of the user library SHOULD be incremented to the same degree:
  - for MAJOR change in transitive dependency we SHOULD increment MAJOR part of our version,
  - for MINOR change in transitive dependency we SHOULD increment MINOR part of our version,
  - for PATCH change in transitive dependency we SHOULD increment PATCH part of our version.
  
If a few dependencies are being changed, only the highest degree is required to be incremented.

aij:
This actually isn't specific to Java, but applies to any system that needs to resolve dependencies to a single version. (Which I think is actually most package managers, with the notable exception of NPM [which will instead wantonly break the build by installing incompatible versions of a dependency *even when they are part of the API*.])

dlow-yahoo-inc:
[...]

PR #414, opened 2017-11-30
Add a note about transitive dependencies on JVM

Primetalk:
Closes https://github.com/semver/semver/issues/341

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -286,6 +286,19 @@ modification depends on whether you updated your dependencies in order to fix
 a bug or introduce new functionality. I would usually expect additional code
 for the latter instance, in which case it's obviously a minor level increment.
 
+#### Special note for JVM or similar environments
+
+In JVM or other environments where only one version of a library could 
+be present in a running system, the transitive dependencies effectively become
+part of the public API and thus the version of the user library SHOULD be 
+incremented to the same degree:
+  - for MAJOR change in transitive dependency we SHOULD increment MAJOR part of our version,
+  - for MINOR change in transitive dependency we SHOULD increment MINOR part of our version,
+  - for PATCH change in transitive dependency we SHOULD increment PATCH part of our version.
+  
+If a few dependencies are being changed, only the highest degree is required to be incremented. 
+(See also [the discussion](https://github.com/semver/semver/issues/341).)
+
 ### What if I inadvertently alter the public API in a way that is not compliant with the version number change (i.e. the code incorrectly introduces a major breaking change in a patch release)?
 
 Use your best judgment. If you have a huge audience that will be drastically
--- a/semver.md
+++ b/semver.md
@@ -288,10 +288,10 @@ for the latter instance, in which case it's obviously a minor level increment.
 
 #### Special note for JVM or similar environments
 
-In JVM or other environments where only one version of a library could 
+In such environments where only one version of a library could 
 be present in a running system, the transitive dependencies effectively become
-part of the public API and thus the version of the user library SHOULD be 
-incremented to the same degree:
+part of the public library interface and thus the version of the user library SHOULD be 
+incremented to the same degree as the version of the transitive dependency:
   - for MAJOR change in transitive dependency we SHOULD increment MAJOR part of our version,
   - for MINOR change in transitive dependency we SHOULD increment MINOR part of our version,
   - for PATCH change in transitive dependency we SHOULD increment PATCH part of our version.


Primetalk:
FYI @jwdonahue

semver.md
```diff
@@ -286,6 +286,19 @@ modification depends on whether you updated your dependencies in order to fix
 a bug or introduce new functionality. I would usually expect additional code
 for the latter instance, in which case it's obviously a minor level increment.
 
+#### Special note for JVM or similar environments
```
aij: That makes it sound like it's a special case when it's actually the norm. It's problematic even in environments that *do* allow multiple versions of the same library.

I think the real issue is that the previous FAQ is misleading.

For example, let's take a really simple API:

```frobnicate: Widget -> Result```

Now upgrade the widget library to an incompatible version. (Eg: 1.0.0 to 2.0.0) Now the API is:

```frobnicate: Widget -> Result```

No change, right? Except this API is incompatible with the previous API because it now requires a v2 Widget instead of a v1 Widget.

Of course anyone reading the previous FAQ *carefully enough* would realize that when the version of a library is *implicitly* part of the public API, then changing it *is* changing the public API despite `diff` not showing a difference. Many people appear to not read it carefully enough.

And also of course, in environments like the JVM that only allow a single version of a library, even when a library is used "purely internally" and not even mentioned in the public API, it is still part of the *undocumented* public API in that it precludes the use of a different version of that library.

File: semver.md [hunk 2]
```diff
@@ -286,6 +286,19 @@ modification depends on whether you updated your dependencies in order to fix
...
+
+In JVM or other environments where only one version of a library could
```
jwdonahue: Just remove "JVM or other", and I think it works.. The SemVer spec implicitly covers what this FAQ change makes explicit.

semver.md
```diff
@@ -286,6 +286,19 @@ modification depends on whether you updated your dependencies in order to fix
 a bug or introduce new functionality. I would usually expect additional code
 for the latter instance, in which case it's obviously a minor level increment.
 
+#### Special note for JVM or similar environments
+
+In JVM or other environments where only one version of a library could 
+be present in a running system, the transitive dependencies effectively become
+part of the public API and thus the version of the user library SHOULD be
```
jwdonahue: I would prefer "Interface" over "API", but that issue is still open to some debate.

Primetalk:
I've created [a PR to `semver.org`](https://github.com/semver/semver.org/pull/180). But I also think that this change might very well be done in `semver` too. It's an important widespread use case.

jwdonahue:
I am currently under the impression that all the active work on the spec is happening over on semver.org. Hence the extreme difference in the depth of the backlogged PR's.  Somebody (@Haacked) might come along and confirm or deny whether this impression is warranted.

jwdonahue:
@Primetalk, @Haacked is the authority on this, my apologies for any confusion.

Primetalk:
@jwdonahue could you clarify what are the remaining obstacles?

jwdonahue:
@Primetalk, Sorry, life has intervened and each time I looked at this, I didn't have time to recover enough context. I may be among the most active on this projects issues section, but I have little influence over what is merged. I think that any accepted changes will necessitate a good deal of work on translations, so this PR may languish until sufficient resources are available.

@Haacked, this looks like a patch level addition to the spec that clarifies what the rules actually mean in an important real-world scenario. It answers a common enough question here and on StackOverflow that  I have come to believe it should be clarified in the FAQ.  I think a strict interpretation of the spec, does imply the transitive relationship between version changes across dependent libraries referenced in a package, but it seems to be open for interpretation. We should consider taking this change or something like it.

fverdino:
@jwdonahue, which translations are mandatory ? I see 22 translations of official page. Who is supposed to provide such translations ? is up to author of Pull Request to provide all this stuff ? before to start translating I guess final confirmation of current English version is needed, not ? 
Last thing:
I know that I could be a little bit offtopic but this translation burden slow down spec evolution. Why this big constraint ? thx

jwdonahue:
Don't worry about it. Actual work here seems to happen in cycles. As far as I know, @Haacked is the only one who can accept PR's and he's a busy guy with a life beyond SemVer.

semver.md
jwdonahue: @aij, you are getting interface and package mixed up. An interface never has any implementation issues to consider.

aij: @jwdonahue Could you elaborate? I didn't think I was talking about packages at all.

I was talking about non-textual (contextual?) interface changes. An interface can be changed in an incompatible way even without any textual changes to it's definition, by merely changing the context. For the above example, if the `Widget` interface is changed incompatibly, it changes the meaning of `Widget -> Result`, even though that text itself is not changed.

The reason I put quotes around "purely internally", is because a lot of environments don't support making use/nonuse of a library into an implementation detail. It necessarily becomes part of the public interface in environments that can't/don't link in multiple copies of the same library. (JVM, *nix shared libraries, most (all?) C static libraries, OCaml, Python, etc)

jwdonahue: @aij, I must have misread something. I thought you were talking about interfaces when in fact you were talking about packages/libraries.

Primetalk:
@jwdonahue The changes you requested in https://github.com/semver/semver/pull/414#pullrequestreview-80298503 has been closed by @Haacked. See comment:
https://github.com/semver/semver.org/pull/180#issuecomment-441395292

jwdonahue:
@Primetalk, Looks likes we have a PR in both semver/semver.org and semver/sember. Guess @Haacked  wants to get this one resolved first. I've never been clear on exactly how changes are supposed to flow into the spec.

alexandrtovmach:

alexandrtovmach:
@Primetalk Thanks for contribution 👍
Could you fix linter warnings?

Primetalk:
@alexandrtovmach - fixed.

grv87:
This note should mention that for JVM a developer could use a package relocation tool, like [shadow](https://github.com/johnrengelman/shadow) in order to avoid all this.

grv87:
Also, the wording is too broad. It's not _JVM environment_, it's just the default class loader.
It's possible to [write a custom isolating class loader](https://www.alibabacloud.com/blog/implementing-java-class-isolation-loading_597343), and some platforms, like [Eclipse](https://wiki.eclipse.org/FAQ_What_is_the_classpath_of_a_plug-in?), have it.

PR #415, opened 2017-12-03
widen 'compatible changes' beyond just bugfixes

### LINKED ISSUE #146: What about refactoring?
Semantic Versioning holds that the patch version should be incremented for...

> backwards-compatible bug fixes

but that... 

> Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version.

Where does refactoring fit into this? Refactored code is modified code, but the modifications do not necessarily contain any bug fixes.

### LINKED ISSUE #339: Some clarifying fixes
I have a bunch of fixes, all clarifying the distinction between the external behaviour (the public API), and the internal implementation. It strikes me that the point of semver is precisely this. We are versioning behaviour, not implementation. At minimum I would suggest:

In the summary about PATCH, replace "backwards compatible bug fixes" with "API identical implementation changes". Backwards compatible is redundant with the previous line, and we may be refactoring or otherwise reimplementing.

Tieske:
fixes #339
fixes #146

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -9,7 +9,7 @@ Given a version number MAJOR.MINOR.PATCH, increment the:
 1. MAJOR version when you make incompatible API changes,
 1. MINOR version when you add functionality in a backwards-compatible
    manner, and
-1. PATCH version when you make backwards-compatible bug fixes.
+1. PATCH version when you make backwards-compatible changes.
 
 Additional labels for pre-release and build metadata are available as extensions
 to the MAJOR.MINOR.PATCH format.
@@ -76,8 +76,8 @@ is incremented after this release is dependent on this public API and how it
 changes.
 
 1. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards
-compatible bug fixes are introduced. A bug fix is defined as an internal
-change that fixes incorrect behavior.
+compatible changes are made. Such changes include but are not limited to:
+internal changes that fix incorrect behavior, or refactorings.
 
 1. Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards
 compatible functionality is introduced to the public API. It MUST be


semver.md
```diff
@@ -9,7 +9,7 @@ Given a version number MAJOR.MINOR.PATCH, increment the:
 1. MAJOR version when you make incompatible API changes,
 1. MINOR version when you add functionality in a backwards-compatible
    manner, and
-1. PATCH version when you make backwards-compatible bug fixes.
+1. PATCH version when you make backwards-compatible changes.
```
Haacked: Under this definition, how would you distinguish MINOR from PATCH?

> MINOR version when you add functionality in a backwards-compatible

Adding functionality is also a "change" so the distinction between MINOR and PATCH is lost with the proposed changes here.

The point of PATCH is that no additional functionality is added. It is only fixes to _existing_ functionality.

semver.md
```diff
@@ -76,8 +76,8 @@ is incremented after this release is dependent on this public API and how it
 changes.
 
 1. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards
-compatible bug fixes are introduced. A bug fix is defined as an internal
-change that fixes incorrect behavior.
+compatible changes are made. Such changes include but are not limited to:
+internal changes that fix incorrect behavior, or refactorings.
```
Haacked: By definition, a refactoring doesn't change the behavior at all. It probably doesn't make sense to release a new public version of a package that _only_ includes refactorings. If you did, it should be the same version since the behavior didn't change.

However, many folks conflate "refactoring" with "bug fixing while cleaning the code" in which case "bug fix" covers the scenario.

semver.md
jwdonahue: @Haacked, good catch, I can't believe I missed that.

How about "PATCH version when you make backwards-compatible changes that do not change the interface"  ?

semver.md
jwdonahue: I do not believe it is possible to refactor code without having an impact on performance or size of the package, either of which carry some risk.  What about security critical code that uses your interface implementation?  Would you want to lie to them?  Is it even legal under SemVer to release different packages with the same name, decorations (arch/debug/codecoverage, etc) and version number on them?  I would hope not.

jwdonahue: One good reason to release a refactored version is to test your assumptions that it is indeed a plug-in replacement for the earlier version and allow some back-time in the consumer community prior to proceeding with new feature development.

jwdonahue: >3. Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version.

Tieske: > By definition, a refactoring doesn't change the behavior at all.

@Haacked I beg to differ, improved performance is a change in behavior. Yet not a "functional" change. A new release based on a refactoring that only improves performance is perfectly valid.

> in which case "bug fix" covers the scenario.

Personally I don't like the terminology of 'bugfix' since it ties the entire spec to software development. Alternative terminology that is more open is preferred imo.

semver.md
Tieske: imo it is very clear, if you ADD functionality, minor, if you change without changing functionality patch.

If you follow the reasoning by @Haacked then there also is no distinction between MAJOR and the others, as MAJOR also entails changes

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API.

Mind the "new" in the above ^^^. 

Those lines at the top are basically the TLDR version of the spec. The details further down are very clear.

semver.md
Tieske: > It probably doesn't make sense to release a new public version of a package that only includes refactorings. If you did, it should be the same version since the behavior didn't change.

@Haacked the purpose of versioning goes beyond just functional changes. From a DevOps perspective I would want each version out there to be separately versioned. A refactoring (though it should not) might  introduce a new bug. And when someone hits that bug I'd want to know in which version of the code that happened.

One might even consider that a bad instruction in the `readme` for getting-started is a reason for a new patch version with the fixed instruction. You probably wouldn't do that, and me neither probably, but it is not up to the SemVer spec to be opinionated on that.

SemVer should not be opinionated, and it is up to the author to specify the exact scope of the version he or she stuck on a release, and there just is no one-size-fits-all for this.

Tieske: That last remark on "scope" should probably become a new FAQ entry.

Tieske:
@Haacked are you inclined to accept a change on this (not necessarily this one)? If not we can close the PR now, otherwise let's discuss how to move forward.

Tieske:
@Haacked ping

semver.md
Haacked: The definition of "refactoring" I use is the one Martin Fowler notes in the title of his book, [Refactoring: Improving the Design of Existing Code](https://amzn.to/2Hf4Mow). In [his bliki](https://martinfowler.com/bliki/DefinitionOfRefactoring.html), he notes,

> Refactoring (noun): a change made to the internal structure of software to make it easier to understand and cheaper to modify without changing its observable behavior.

Thus any change in behavior including performance improvements would not be a refactoring.

> either of which carry some risk

This resonates with me. It doesn't make sense to me that you'd ship a release with just refactorings (as defined by Fowler) because there shouldn't be any outwardly visible change. But my preference doesn't stop people from doing that. And if they do, there's always the risk they've incorrectly refactored the code and introduced a change.

Semver is about communicating risk. So it make sense to me to release a patch update in this case.

However, still not convinced by the specific proposed language. As noted elsewhere, you can add a new feature and it would still be a backward compatible change.

I think the spirit of the patch is that there's no outward functional change other than behavior is corrected (bug fix) or performance is improved. Are there other backward compatible changes that don't fall in that category?

semver.md
jwdonahue: How about this?

MAJOR version when you make incompatible API changes,
MINOR version when you add to the API in a backwards-compatible manner, and
PATCH version when you make any other backwards to compatible changes.

semver.md
jwdonahue: How about this?

MAJOR version when you make incompatible API changes,
MINOR version when you add to the API in a backwards-compatible manner, and
PATCH version when you make any other backwards compatible changes.

jwdonahue: And:

6. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible changes are introduced. This includes internal changes that fix incorrect behavior (bug fixes), but does not include any changes to the public API.

jwdonahue: @Haacked, Fowler is wrong in the strictest sense. It is not possible to modify source code, or the compiler for that matter, without making "observable changes". He's correct in the sense that the intended behavior, it's functionality, is not changed by refactoring. We can argue over whether a few nano-seconds here or there really constitutes a perf gain or not, but I have worked on systems where such differences are flagged by automation, trigger broader longer running test runs and must be signed-off on.  For some security conscious consumers, even a change in the content hash of the bits is reason for alarm.

Tieske: @jwdonahue 

> It is not possible to modify source code, or the compiler for that matter, without making "observable changes".

I don't think Fowler is wrong (if even possible at all, since definitions are almost always opinionated). How about replacing an `if-then` sequence by a `select`? "_to make it easier to understand and cheaper to modify_".

jwdonahue:
@Tieske, only slightly off topic... you might be interested in taking a look at [VersionMeta](https://www.versionmeta.org/). Still a work in progress, but the ultimate goal is to provide the means for publishers of any content type to define the actual semantics of the version string(s) they deploy. So you SemVer might be 1.0.0 {SemVer.2.0.0} while a text-book might have 1.3 {https://books.org/Simple.xml}. Either one would be machine readable by the same library of code, define precedence rules, semantics of defined fields and provide links to human readable specs. 

Of particular interest for adopters of SemVer, they could refine their VersionSchema to include details about each of the prerelease and build meta identifiers (fields) as well.  I for one would love to be able to scrap the current precedence rules on prerelease tags in favor of an ordered list of enumerators (Experimental, Prototype, Flight#, RC#, etc.  If the weather doesn't doesn't get too nice, I should have an early preview of [VersionSchema](http://versionschema.org) posted sometime in the next week or two.

semver.md
Tieske: @jwdonahue from a previous comment: if in a repo the docs are updated because of some faulty instruction in those docs. Some would consider it a bugfix, some wouldn't. It's up to the package owner to define whether the docs are in scope of the SemVer version number applied to the package. Or maybe he could choose to have 2 version numbers, 1 for the code, 1 for the docs.

Another example (to stay away from zebra's and books 😄): a components has a REST interface, but it also stores some data in a DB. Some users tend to directly access the DB besides the REST interface. Now if a SemVer version is applied, is that on the REST interface, or does that include the DB schema?

It's always up to the author of a package/component to define what is in scope and what not.

Tieske:
> @Haacked are you inclined to accept a change on this (not necessarily this one)? If not we can close the PR now, otherwise let's discuss how to move forward.

ping again

semver.md
levitte: You could very well make backward compatible changes that would merit a PATCH level release:

``` diff
-int hash(char *str);
+int hash(const char *str);
```

For all intents and purposes, this change doesn't do anything other than to openly declare that `hash` should do what it probably already doesn't do, i.e. it doesn't change the contents pointed at by `str`, and now says so clearly.

This change, however, merits a MAJOR release, as this is an *incompatible* change:

``` diff
-int hash(const char *str);
+uint64_t hash(const char *str);
```

Tieske:
@haacked ping

Tieske:
@haacked ping

FichteFoll:
The maintainers of semver changed and haacked@ isn't involved anymore, so pinging isn't of any use.

jwdonahue:
@Tieske 

Well I'd say this PR has near zero chance of acceptance at this point.  If you wish to continue promoting it, you'll have to resolve all the conflicts, otherwise, please close it at your earliest possible convenience.

Issue #189, opened 2014-03-27
Define "package"


ecki:
In the Java and OSGi world "package" is a term with a specific meaning. (And especially in OSGi you could even have a version number on seperate packages inside a bundle). I would therefore recommend you define that with package you actually mean "deliverable" or "deploymet unit" and not "java package". Or you define that the granularity at which you apply the version number might be on "API", "(osgi) package", "packages", "operation" or "deployment unit" or "application" level.

jwdonahue:
@ecki, I have always assumed that when a specification does not provide its own definition of a word, the "standard dictionary" for the language the spec is written in, provides the definition.  In this case try looking at http://www.dictionary.com/browse/package?s=t and assume the SemVer spec intended a liberal application of those definitions, or to quote @mojombo's own words from the introduction:

> I call this system "Semantic Versioning." Under this scheme, version numbers and the way they change convey meaning about the underlying code and what has been modified from one version to the next.

Unless you have further questions or wish to issue a PR, please close this issue at your earliest possible convenience.

ecki:
I dont agree that the dictionary link helps. Especially not since the **programming language** the spec applies to does define `package` differently. But sure I can open a pull request.

jwdonahue:
Now that's are a really good start! I think you should issue a PR to semver/semver.

@Haacked (FYI).

ecki:
Looks like I am the only one struggling with the definition.

PR #416, opened 2017-12-06
FAQ: what is a package

ecki:
Clarify that package talks about deployment units.
Introduce the idea that it can apply to other constructs as well.
Add Java as example technology since it has different terminology.

This fixes semver#189

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -319,6 +319,24 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
+### What does SemVer consider a package?
+
+With "package" SemVer describes deployment units, independent of the actual
+technology used. Packages could be built from source artifacts or assemble
+binary artifacts.
+
+An actual ecosystem might have a different terminology. For example In *Java*
+a typical deployment unit would be a *JAR file*. On the other hand a *Java
+package* is typically not independently versioned. When using *OSGi bundles*
+a *Java package* is a manageable unit and Semantic Versioning supports the
+change lifecycle.
+
+### Does SemVer apply to APIs?
+
+Web APIs, Specifications, logical collections of programming interfaces
+or implementations might not consist of actual deployment units, however
+Semantic Versioning can still be used to describe the impact of changes.
+
 
 About
 -----


ecki:
Should I raise the version of the document to 2.0.1 in this PR or keep that seperate?

FichteFoll:
It doesn't touch the actual spec, so no imo.

Regardless, bumping the version should only occur when a release is made by updating semver.org since you wouldn't know what kind of changes would eventually end up in that release until it is frozen.

semver.md
```diff
@@ -319,6 +319,22 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
+### What does SemVer consider a package?
+
+The meaning of package in this specification describes a deployment unit.
+It could be built from a single or multiple source artifacts. But it can
```
jwdonahue: > ... artifacts. But it can...

Should either be "...artifacts, but it can..." or "...artifacts. It can..."

semver.md
```diff
@@ -319,6 +319,22 @@ Abbreviating "version" as "v" is often seen with version control. Example:
...
+also be an assembly of multiple binary artifacts.
+
```
jwdonahue: No need for one sentence paragraphs here.  Just connect the above to the one below and then the "however" in that lower sentence will not be cognitively burdensome.

semver.md
```diff
@@ -319,6 +319,22 @@ Abbreviating "version" as "v" is often seen with version control. Example:
...
+Semantic versioning concepts can however also be applied to logical
+collections like an interface, an API or any other concept which
+undergoes changes and can be assigned a version.
+
+In Java for example a typical deployment unit would be a JAR file and a
+Java package is typically not versioned independenly.
+However with modular technologies (like *OSGi bundles*) the Java package
```
jwdonahue: Very minor nit: this sentence should start on the line above.

jwdonahue:
I started a review with some comments.  Thank you @ecki for contributing this, I hope it is accepted.  When it is, please don't forget to close #189.

semver.md
ecki: Do you mean breaking the line after ‚However with...‘? 

(i can do that, I am just used to breaking lines by sentence for markup, which makes generally more stable diffs. But I will try to fit in the existing style)

semver.md
ecki: Yes correct, was doing too much Asciidoc lately. And I will remove ‚the meaning‘

semver.md
ecki: Correction: will line join it as well

jwdonahue: > "which makes generally more stable diffs"

Oh, I hadn't thought of that. Like I said, it was a nit, perhaps you should ignore me on this one.

semver.md
jwdonahue: "..." means whatever came before or whatever comes after. No need to insert them in the text.

semver.md
FichteFoll: Relevant: http://rhodesmill.org/brandon/2012/one-sentence-per-line/

FichteFoll:
It will be closed automatically because the phrase "fixes #189" is included in the OP.

ecki:
I have updated the Pull Request. I use the already existing formatting (i.e. not use line-per-sentence format). Please let me know what you think now.

semver.md
```diff
@@ -319,6 +319,24 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
+### What does SemVer consider a package?
+
+With "package" SemVer describes deployment units, independent of the actual
+technology used. Packages could be built from source artifacts or assemble
+binary artifacts.
+
+An actual eco-system might have a different terminology. For example In *Java*
```
jwdonahue: [Ecosystem](http://www.dictionary.com/browse/ecosystem?s=t) is one word, no need for the hyphen.

ecki: Thanks @jwdonahue  - pushed a new version.

jwdonahue: Every morning I seem to see something new. Let's accumulate some comments for a few days and see where it lands.  I think we're almost there.  The second paragraph looks fine to me, but I nearly dislocated a cog on the last sentence in the first paragraph. Either the paragraph needs another sentence or ... I don't know yet.  I need to let it ferment.

sander1095:
This has been dead for a while, commenting to revive :)

PR #420, opened 2017-12-16
Create a LICENSE file.

jwdonahue:
Should be SOP for all GitHub projects to have a license.   The unlicense seems appropriate for the SemVer spec.


jwdonahue:
Woops!  I see there is a creative commons license referenced from the spec.  This should be in a license file, I'll update the PR...

PR #437, opened 2018-03-31
Fix link to issues

guillermocalvo:


jwdonahue:
@Haacked, this is simple and appears to be correct. https://github.com/mojombo/semver/issues currently redirects to https://github.com/semver/semver/issues anyway.

hartwork:
Go?

Haacked:
Thanks! Good call!

PR #451, opened 2018-07-25
Stylise "backwards compatible" consistently

ExplodingCabbage:
Later in the document, it's consistently spelt with a space, rather than a hyphen - so I figure it should be here, too.


hartwork:
I have checked `git grep -C1 compatible | grep 'compatible\|backwards'` output and conclude: Right direction, good patch. :+1:

Haacked:
I like the consistency. Thanks!

PR #452, opened 2018-07-29
Clarify if change is breaking for a specific platform only

### LINKED ISSUE #444: SemVer declares API changes, but not the minimal required platform/engine/environment/framework
Please take a look at this issue: https://github.com/request/request/issues/2772#issuecomment-330948055

TL;DR version:
* API did not change, so the software was releasing only the MINOR.
* The software minimal required platform (in this case node.js engine version) did change from 0, to 4, to 6.
* As the result the software could not be run on the unsupported platforms (node.js v0, v4, v6). It crashes production servers at runtime for many people out there causing revenue loss.

-----

General best practice in node.js community:
* Increase the MAJOR version when your software minimal platform version increases.
(E.g. if minimum supported node.js increases from v4 to v6 you bump from v2.86 to v3.0.)

The above mentioned issue practice:
* Follow SemVer. Meaning: no API changes -> MINOR version upgrade.

-----

Proposal:

* Either mention that "bumping the minimal required platform should be considered a MAJOR release". Or
* "bumping minimal required platform should not increase the software MAJOR version".

-----

Please note, this is a serious issue. Please, avoid platform-specific holy wars. Let's improve SemVer even further.

koresar:
As discussed in #444 with @jwdonahue, I'm supplying a PR which clearly states that even if API does not change but a software update breaks something then it should be a MAJOR release.

Closes #444

Probably this is not the best wording ever, and maybe even contradicts something else, but I took a careful look over the the whole specification. I believe this change is compatible with everything else in the specification.

If anything should be reworked - reworded - removed - rethought - etc, I'm happy to continue.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -6,7 +6,7 @@ Summary
 
 Given a version number MAJOR.MINOR.PATCH, increment the:
 
-1. MAJOR version when you make incompatible API changes,
+1. MAJOR version when you make incompatible changes,
 1. MINOR version when you add functionality in a backwards-compatible
    manner, and
 1. PATCH version when you make backwards-compatible bug fixes.
--- a/semver.md
+++ b/semver.md
@@ -87,9 +87,11 @@ within the private code. It MAY include patch level changes. Patch version
 MUST be reset to 0 when minor version is incremented.
 
 1. Major version X (X.y.z | X > 0) MUST be incremented if any backwards
-incompatible changes are introduced to the public API. It MAY also include minor
-and patch level changes. Patch and minor version MUST be reset to 0 when major
-version is incremented.
+incompatible changes are introduced to the public API. Major version MUST also
+be incremented if the software can no longer run on any of the
+platforms or frameworks it used to run on, even if API stayed the same. 
+It MAY also include minor and patch level changes. 
+Patch and minor version MUST be reset to 0 when major version is incremented.
 
 1. A pre-release version MAY be denoted by appending a hyphen and a
 series of dot separated identifiers immediately following the patch


semver.md
```diff
@@ -6,7 +6,7 @@ Summary
 
 Given a version number MAJOR.MINOR.PATCH, increment the:
 
-1. MAJOR version when you make incompatible API changes,
+1. MAJOR version when you make incompatible changes,
```
jwdonahue: Rather than remove API, why not also include package?  I think that in the case of NPM and other package managers, the defined dependency tree is part of the package API, but that might be stretching it a bit.

... when you make incompatible package or API changes,

semver.md
```diff
@@ -87,9 +87,11 @@ within the private code. It MAY include patch level changes. Patch version
 MUST be reset to 0 when minor version is incremented.
 
 1. Major version X (X.y.z | X > 0) MUST be incremented if any backwards
-incompatible changes are introduced to the public API. It MAY also include minor
-and patch level changes. Patch and minor version MUST be reset to 0 when major
-version is incremented.
+incompatible changes are introduced to the public API. Major version MUST also
+be incremented if the software can no longer run on any of the
```
jwdonahue: Again, I think the only change should be "...package or public API".

semver.md
koresar:
```
1. MAJOR version when you make incompatible package or API changes,
```
I like how it sounds.

koresar:
Thanks! I'll go through the whole file this week.

koresar:
Yeah. I believe that "incompatible changes" is the best wording. And further down we should explain what the "incompatible changes" mean.

The "package changes" is too broad.

jwdonahue:
@koresar, you'll have to resolve conflicts for this PR to ever get accepted.

koresar:
Thanks for poking mate.

I'll create a new very minimal PR.
Single change: will replace `API`  with `package or API` as you suggested.

koresar:
@jwdonahue Please see #546
Thank you

Issue #232, opened 2014-11-13
RegEx for validating SemVer-numbers


DavidFichtmueller:
I just created a RegEx to check if a version number is a valid Semantic Version Number according to the specification. 

```
^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$
```

Maybe this is useful for some people out there or it could be added to the documentation.

hftmarkand:
Would be nice to have this in the official SemVer specification !

BartG95:
If I'm not mistaken 
````
\d*[a-zA-Z-][0-9a-zA-Z-]*
````
is the same as
````
[0-9a-zA-Z-]+
````
(note the plus)

quantumpacket:
No, the first one requires that there be at least one character from the range `[a-zA-Z- ]` since it has no wildcard. However, the second one does not require at least one char from that range. The string `12345` fails the first, while it passes the second.

BartG95:
Right, but it is part of
````
0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*
````
So "12345" wil pass anyway.

chepner:
If this isn't part of an actual implementation, it should use only (POSIX) standard features, like `[[:digit:]]` in place of `\d`.

quantumpacket:
> If this isn't part of an actual implementation, it should use only (POSIX) standard features, like [[:digit:]] in place of \d.

That's the problem with using a regex for a spec. There are different flavors of regex, which does not make it really a standard then. Preferably we should have the [ABNF (RFC 5234)](https://tools.ietf.org/html/rfc5234) grammar in the spec for proper validation purposes.

silkentrance:
Perhaps you should get into touch with gulp-bump and ask them for using the official regex:

https://github.com/stevelacy/bump-regex/blob/f70505708719e2f9b7df44e0521145907349b761/index.js#L28

silkentrance:
There is also https://github.com/sindresorhus/semver-regex, perhaps you want to join efforts here.

silkentrance:
And there is also https://github.com/semver/semver.org/issues/59. Get your gears straight.

[...]

PR #460, opened 2018-08-23
FAQ: semver regex

wolf99:
With regex by @DavidFichtmueller taken from https://github.com/semver/semver/issues/232

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -327,7 +327,7 @@ There are two. One with named groups for those systems that support them:
 ^(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-(?<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
 ```
 
-And one with non-capturing groups instead (so cg1 = major, cg2 = minor,
+And one with numbered capture groups instead (so cg1 = major, cg2 = minor,
 cg3 = patch, cg4 = prerelease and cg5 = buildmetadata):
 
 ```
--- a/semver.md
+++ b/semver.md
@@ -319,6 +319,20 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
+### Is there an efficient regex?
+
+There are two. One with named groups for those systems that support them:
+
+```
+^(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-(?<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
+```
+
+And one with non-capturing groups instead (so cg1 = major, cg2 = minor,
+cg3 = patch, cg4 = prerelease and cg5 = buildmetadata):
+
+```
+^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
+```
 
 About
 -----
--- a/semver.md
+++ b/semver.md
@@ -319,7 +319,7 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
-### Is there an efficient regex?
+### Is there a regular expression (RegEx) to check a SemVer string?
 
 There are two. One with named groups for those systems that support them:
 
--- a/semver.md
+++ b/semver.md
@@ -324,7 +324,7 @@ name and the semantic version is "1.2.3".
 There are two. One with named groups for those systems that support them:
 
 ```
-^(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-(?<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
+^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
 ```
 
 And one with numbered capture groups instead (so cg1 = major, cg2 = minor,
@@ -333,6 +333,10 @@ cg3 = patch, cg4 = prerelease and cg5 = buildmetadata):
 ```
 ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
 ```
+
+Both of these expressions are compatible with ECMA Script (JavaScript), PCRE
+(Perl compatible regular expression), Python and Go.
+
 
 About
 -----
--- a/semver.md
+++ b/semver.md
@@ -319,7 +319,7 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
-### Is there a regular expression (RegEx) to check a SemVer string?
+### Is there a suggested egular expression (RegEx) to check a SemVer string?
 
 There are two. One with named groups for those systems that support them (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python and Go).
 
--- a/semver.md
+++ b/semver.md
@@ -319,7 +319,7 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
-### Is there a suggested egular expression (RegEx) to check a SemVer string?
+### Is there a suggested regular expression (RegEx) to check a SemVer string?
 
 There are two. One with named groups for those systems that support them (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python and Go).
 
--- a/semver.md
+++ b/semver.md
@@ -321,21 +321,22 @@ name and the semantic version is "1.2.3".
 
 ### Is there a regular expression (RegEx) to check a SemVer string?
 
-There are two. One with named groups for those systems that support them:
+There are two. One with named groups for those systems that support them (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python and Go).
+
+See https://regex101.com/r/Ly7O1x/3/
 
 ```
 ^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
 ```
 
 And one with numbered capture groups instead (so cg1 = major, cg2 = minor,
-cg3 = patch, cg4 = prerelease and cg5 = buildmetadata):
+cg3 = patch, cg4 = prerelease and cg5 = buildmetadata) that is compatible with ECMA Script (JavaScript), PCRE (Perl Compatible Regular Expressions, i.e. Perl, PHP and R), Python and Go.
+
+See https://regex101.com/r/vkijKf/1/
 
 ```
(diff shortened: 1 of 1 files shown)


wolf99:
Just wanted to get this kicked off really - additions welcome!

DavidFichtmueller:
Thanks for creating the PR. I have had it on my list, but didn't had the time for it, as it was a lower priority than a lot of other stuff. 

I think the headline could be a bit more descriptive, maybe something like "Is there a regular expression (RegEx) to check a SemVer string?" but I am also ok with the version as it is right now.

semver.md
```diff
@@ -319,6 +319,20 @@ Abbreviating "version" as "v" is often seen with version control. Example:
...
 
+### Is there a regular expression (RegEx) to check a SemVer string?
+
+There are two. One with named groups for those systems that support them:
+
+```
+^(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-(?<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
+```
+
+And one with non-capturing groups instead (so cg1 = major, cg2 = minor,
```
jwdonahue: These both have capture groups, the second one has numbered capture groups, commonly referred to as just "capture groups", the first one has "named capture groups". A non-capturing regex would only be useful as a validity test, either the string matches or it doesn't.

DavidFichtmueller: Well both have non-capturing groups as well. They are for groupings that are structurally necessary (e.g. in combination with qualifiers) but should not be referenceable as a numbered capture group. 

But you are correct that this description of the second regex is a bit misleading. I also haven't noticed this before. How about "And one where the numbered regex groups correspond to the segments of SemVer (so cg1 = major, ..."

wolf99: Would "with unnamed capture groups" or "with non-named capture groups" be more succinct?

wolf99: Would "And one with numbered capture groups instead" be more succinct?
The parenthesised information remains - I guess this assumes a tiny bit higher level of understanding of regexs, but do we otherwise start explaining capture groups?

DavidFichtmueller: Yes, I think this is sufficient. The people who will want to know this, will most likely have a general understanding of RegExs and capture groups. Giving a short overview which capture group returns which segment of the SemVer is what they might want to know at hand. This allows the usage of the RegEx without fully understanding the details (non capturing groups, etc.).

jwdonahue: Is it really common that an implementation supports capture groups and not named capture groups?

wolf99: @jwdonahue from a quick Google it seems older versions of some languages do not support them (e.g Java < v7) and even some not so old (ECMAScript < 2018), given that older is relative to one's requirements (Java v7 was only released a gnats eye-blink ago in terms of longer lived applications), then I'd opine that yes, it is common 😃

jwdonahue: @wolf99, interesting. Well, I think the FAQ should refer to two regex forms for capture groups, named and otherwise, and then there should be a non-capturing version for simple validity testing. Not every step in a DevOps process cares a gnat about any of the fields, as long as the string is well formed.

DavidFichtmueller: @jwdonahue well the non-named groups are just parentheses which are supported by pretty much every system except POSIX/GNU BRE/ERE according to [this site](https://www.regular-expressions.info/refcapture.html). One could argue for making one that doesn't use non-capturing groups, so the reference numbers will be different, but then again, according to the same site, this only effects ORACLE and XML. So I would just leave it as it is. In case some people raise this question at some point (which I highly doubt), we can further extend the FAQ. But leave it concise for now.

wolf99:
Are you folks OK with this as it currently is? 
If so, if we could get it approved and merged would be great
If not, any specific areas of change?

hartwork:
It would be nice to mention which "flavour" of regex syntax is used. It looks like made for Python to me but that's only a quick guess and would nice to be communicated. http://www.greenend.org.uk/rjk/tech/regexp.html gives examples of differences in syntax flavors, not just regex engine features.

chriskilding:
My team was interested in using this regex for a devops utility, but we wanted an automated way to prove the regex was solid. So we wrapped it in a small script and wrote automated tests using [Bats](https://github.com/bats-core/bats-core).

This was a success, and we realised that since we already had an automated suite, we might as well have a 'reference implementation' to go with it. So we expanded the script around the regex into a full program with all the features we needed.

We also implemented - and cleaned up - the extended validation suite from https://github.com/semver/semver/issues/232#issuecomment-430813095 to fully battle-test the validation use case.

The utility is now open source and available at: https://github.com/chriskilding/semver

You can see the regex inside it at: https://github.com/chriskilding/semver/blob/master/semver#L14

We suspect that either the `semver` program or the Bats test suite will be of interest in this pull request.

We have also realised that **a second regex is required for precedence comparisons**, but it has not been provided yet in this pull request. That regex should extract all relevant identifiers in the prerelease component so that a 'greater than' or 'less than' comparison can be made.

chriskilding:
Here’s a first crack at the regex for semver precedence comparison. It’s not ideal, so I’m hoping a regex guru can help improve it...

We experienced problems with capturing repeated matches of a capture group. Our workaround was to capture the first pre-release identifier separately from the rest.

In pseudocode our recipe looks like:

    version = ‘1.2.3-prerelease+build’
    prerelease = get-prerelease(version)
    head = match(prerelease, head-regex)
    tail = match(prerelease, tail-regex)
    identifiers = concat(head, tail)

The regexes are:

- head-regex: `/^([0-9a-zA-Z-]*)/`
- tail-regex: `/\.(0|[1-9a-zA-Z-][0-9a-zA-Z-]*)/g`

bradcupit:
@wolf99 @DavidFichtmueller 
The team I'm on would like to use this regex, but since the PR hasn't merged the license is ambiguous. Would it be safe to assume the license is [CC BY 3.0](https://creativecommons.org/licenses/by/3.0/) since that's [the license of semver](https://github.com/semver/semver/blob/master/semver.md#L337)?

DavidFichtmueller:
@bradcupit yes, I see it the same way, that the content of this PR is under the same license as the site it will be merged into (hence CC BY 3.0) but the BY clause would apply to semver.org as the source. Since the PR is not yet merged you might want to link here as well just to be sure, so people who will later look into this code understand where the regex came from. Hope this helps.

jwdonahue:
@wolf99, https://help.github.com/en/articles/github-terms-of-service#d-user-generated-content  applies.  

The regex contributor(s) retain ownership of their own works, but grant license equivalent to SemVer's CC BY 3.0, by virtue of posting it here.

wolf99:
The above comments by @jwdonahue and @DavidFichtmueller should indeed cover any eventuality, but just to round out the question, I understood when making this PR that the contribution, merged or not, would be covered by the same license used by this repository.

wolf99:
In answer to the question from @hartwork on regex flavour; I actually am unsure / do not remember the details. The proposed regex is the culmination of the work on the previously linked issues - particularly @DavidFichtmueller. Maybe David can comment?

DavidFichtmueller:
@hartwork I wrote and test the regex as ECMA Script (Java Script) and later PCRE (Perl compatible regular expression) for the variant with the named capture groups. But now I had a closer look and realized while this was compatible to Perl, it wasn't to Python because their syntax for named capture groups is a little different. With a little variation it is however possible to make it compatible to both Perl and Python (and Golang as well, all of it as tested on https://regex101.com/r/Ly7O1x/1/ ). Here is the updated version: 
`^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
Notice the additional capitalized P in front of each capture group name. @wolf99 can you update this expression in another commit to this PR? 
Also the version without the named capture groups works for all tested flavors (ECMA, Perl, Python, Golang).

wolf99:
@DavidFichtmueller Rebased and added 😺

DavidFichtmueller:
@wolf99 Thanks, but I think I didn't make myself clear. Only the one without the named capture groups works with JavaScript (ECMA) but this will work on all other of the tested languages as well. For the named capture group it now works with Perl, Python and Go (whereas previously it was only with Perl) but not JavaScript. 
Maybe we could add this as: 

> One with named groups for those systems that support them (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python and Go):

and later 

> [RegEx without named capture groups]
> This expressions is compatible with ECMA Script (JavaScript), PCRE (Perl Compatible Regular Expressions, i.e. Perl, PHP and R), Python and Go.

Thanks and sorry for the confusion.

DavidFichtmueller:
As for the stored version at regex101, I don't really know how long they are stored but I would assume this is for as long as the service is around. Linking to it is a good idea, maybe right after the expression a little link "(open on regex101)". 
I have updated the description of the test string a little bit, so you can now link to https://regex101.com/r/Ly7O1x/3/ for the named capture groups and to https://regex101.com/r/vkijKf/1/ for the unnamed capture groups. 
In case the site eventually stops working or the versions aren't stored anymore, we can just remove the links from the MD.

wolf99:
Thanks for the clarification @DavidFichtmueller, have updated.

DavidFichtmueller:
Thanks. @jwdonahue can this PR be merged now?

jwdonahue:
@DavidFichtmueller, it's never been up to me.

DavidFichtmueller:
@jwdonahue Oh, my bad. Sorry for the mix up. @haacked can this me merged or does this need to go through the RFC process outlined in CONTRIBUTING.md as well? From my understanding this wouldn't be necessary as this just adds an FAQ which doesn't effect the interpretation of SemVer.

haacked:
No [longer up to me either](https://haacked.com/archive/2019/02/11/semver-collective/). Maybe @isaacs has an opinion.

isaacs:
For what it's worth, this is the regexp that node-semver uses:

```
> semver.re[semver.tokens.FULL]
/^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][a-zA-Z0-9-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][a-zA-Z0-9-]*))*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$/
```

DavidFichtmueller:
@isaacs this regex is exactly the same as the one with the non-named capture group from the PR with the exception that it allows for a "v" to precede the version number, which is not allowed by the standard. I can understand that when evaluating the code in node it can be a bit more flexible in this regard but for the FAQ entry it should be consistent with the rest of the document.

isaacs:
@DavidFichtmueller Cool! Yeah, we probably don't want the `v` to be included in the spec, but it's so common in practice that we just allow it in node-semver.

wolf99:
Are you able to merge then @isaacs ?
If not, it looks like those listed in semver / people are @anangaur, @segiddins and @steveklabnik - can you merge?
Thanks.

steveklabnik:
I am not entirely sure that I want to promote a particular implementation, of which this would be.

If other people on the team really want this, then I won't stop it, but I'm not sure I'm really for it either, to be honest.

isaacs:
Yeah, I'm generally wary of reference implementations, as a rule.

I'm sure @steveklabnik will agree that both the `semver` crate and `semver` npm package are amazing and created by developers who are truly fantastic people (lol), and this regexp matches node-semver exactly.  But if there's a bug in my code (and it wouldn't be the first time by a wide margin), then matching it means that there may be a bug in the spec.  And spec bugs are way more difficult to fix than bugs where the implementation deviates from the spec.

That being said, I think there _is_  an EBNF representation of the spec, which isn't far off from a regular expression representation.  Maybe this isn't so different, idk.  But I'm not sure what it adds, either.  If you want implementation examples, there are plenty to be found.

So yeah.  I'm also a very weak -0 vote on this.  If someone else on the team wants it badly, I won't block it, but I also don't really wanna vote for it.

wolf99:
@isaacs Not sure where the mention of crate or npm has crept in from? 

This PR is simply to provide a suggested suitable regex to those that are looking to incorporate use of semver into their system or process. It is prompted by me having this very need. I know nothing about regex or npm (never mind whatever 'crate' is?), I simply wanted to find something that would allow me to have sensible versioning in my process. The issue associated to this PR (#232, among others) informs that there are many others in similar positions. I have adjusted the question in the document to indicate that the regexs provided are suggested only (i.e. not canonical).

If you decide that you do not want a FAQ question that helps people that want a regex for incorporating semver, _then please close this PR_. Otherwise please indicate what improvements would lead to a merge (afraid I'm unfamiliar with EBNF so am unsure if your mention of it is a suggestion in this regard or not, if it is, please feel free to submit your own changes in this regard).

isaacs:
@wolf99 The mention of crate and npm is somewhat tongue in cheek, sorry for the in-joke. (Crate is rust's package manager.  @steveklabnik is the maintainer of crate's semver implementation, and I maintain npm's, which both implement basaically this exact regexp.)

EBNF is extended backus-naur form.  Ie, this: https://github.com/semver/semver/blob/master/semver.md#backusnaur-form-grammar-for-valid-semver-versions

The reason why I'm torn on it (and I think @steveklabnik is as well) is that a regexp is kind of on the border between "a formal reference" and "an implementation".  Since this expression includes named capture groups, it's using features that are not included in many popular regexp implementations, making it somewhat platform-specific.

I think that you bring up a valid need, though.  Perhaps we could thread this needle by including a list of known-pretty-good (or at least, popular) SemVer implementations, and an appendix of regular expressions that work on a few different known regexp platforms?

jwdonahue:
Perhaps the best reason to post this regex as an example, is to counter-act all of the incorrect regex examples that can be found by the hundreds.  At least this one has been independently developed by several implementers over the years and its been tested and properly reviewed by people who understand what the SemVer spec actually says.

isaacs:
@jwdonahue Yeah, that is a really good point.  K, I'm over the line on this, then.

@semver/maintainers I intend to land this in 48 hours.  Speak now or hold your peas :)

semver.md
```diff
@@ -319,6 +319,25 @@ Abbreviating "version" as "v" is often seen with version control. Example:
 `git tag v1.2.3 -m "Release version 1.2.3"`, in which case "v1.2.3" is a tag
 name and the semantic version is "1.2.3".
 
+### Is there a suggested egular expression (RegEx) to check a SemVer string?
```
indirect: Should "egular" be "regular"?

ljharb: i wanted to comment `r?` but i thought it’d be too subtle

isaacs: @ljharb I get this joke.

wolf99: Pushed a commit to +r

isaacs:
Alrighty!  Landed in the spec, and pushed the updated spec to the website as well.

https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

Thanks!

joewiz:
@DavidFichtmueller @jwdonahue @wolf99 @isaacs Thanks so much! This regex was really helpful in helping the eXist community create a regex parser - https://github.com/eXist-db/semver.xq/blob/master/src/main/xquery/semver.xq#L43-L47.

MyNameIsCosmo:
For future reference, here is a POSIX-compliant version of the expression:
```
^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)){0,1}(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)){0,1}$
```

gvlx:
> (0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)){0,1}(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)){0,1}

I'm sorry @MyNameIsCosmo , your version is failing on some cases, please recheck: https://regex101.com/r/vHmg9U/1

MyNameIsCosmo:
@gvlx - I believe this is due to missing the beginning of line (`^`) and end of line (`$`) characters:
https://regex101.com/r/vHmg9U/2

Here is the same regex of the named capture group without the characters:
https://regex101.com/r/Ly7O1x/31

Issue #465, opened 2018-10-04
How many hypens are you allowed?


elduddz:
I am trying to get clarification on how many hyphens you can have in the prerelease tag?

I have noted wording in https://semver.org/spec/v2.0.0.html#spec-item-9

A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.

From this I have assumed you can have only one hyphen.

Someone then pointed me to a previous issue raised: https://github.com/semver/semver/issues/78#issuecomment-14981739

if you are allowed more than one hyphen, why is the wording hyphen and not hyphens? and there is nothing in the examples with more than 1 hyphen.

klehelley:
The first hyphen is not actually part of the pre-release part of the version identifier, but the separator between the X.Y.Z part and the pre-release part. The pre-release part can contain hyphens. The way I understand it, that means that when evaluated, version identifier `1.0.2-0.pre-prod.1-2+abcde` will be interpreted this way:

* X.Y.Z part:
  * MAJOR: `1`
  * MINOR: `0`
  * PATCH: `2`
* Pre-release part:
  * 1st identifier: `0` (numeric)
  * 2nd identifier: `pre-prod` (string)
  * 3rd identifier: `1-2` (string)
* Build metadata part: `abcde`

You are right I think however to point out that to match the provided regex and the pluralization of "ASCII alphanumerics", the second instance of "hyphen" in the quoted text should probably be pluralized.

jwdonahue:
@elduddz, good catch! But the issue isn't why isn't it plural, the issue is that it's not plural.  See #460 for a PR with links to related discussion on the actual regex that is proposed to be added to the FAQ. In the one or more of the related discussions you will find a largish test-data set that includes valid semver strings similar to 1.2.3----RC-SNAPSHOT.12.9.1--.12. We should update the spec to say something like:

> MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-].

[...]

PR #466, opened 2018-10-08
Hyphen should be hyphens in the pre-release clause

jwdonahue:
Fix for #465 How many hypens are you allowed in the prerelease tag.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -104,7 +104,7 @@ normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7,
 
 1. Build metadata MAY be denoted by appending a plus sign and a series of dot
 separated identifiers immediately following the patch or pre-release version.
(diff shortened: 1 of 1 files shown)


jwdonahue:
Updated.

r00ster91:
Can any maintainer approve this? I would like to clearly know if it's valid to use multiple hyphens or not.

elduddz:
@r00ster91 the answer is yes, but agree the doc needs that clarification.

jwdonahue:
@r00ster91, see the threads referenced in #460, regarding regex. It's long been well established that `---` is a valid prerelease tag consisting of two hyphens. It is unlikely that you will find anyone to argue otherwise.

jwdonahue:
@anangaur, @dherman, @indirect, @isaacs, @segiddins, @steveklabnik, @Seldaek,

Do any of you have any objections to committing this trivial update?

PR #492, opened 2019-02-11
Spell out the list of maintainers

Haacked:
The @semver/maintainers team member list is not publicly displayed even though the team is public. It's only public to members of the org. This PR spells out the list of members and their affiliations.


Haacked:
I will YOLO-merge this as it's not a change in the spec but a correction to the `CONTRIBUTING.md` and is important too the announcement of the new SemVer owners.

From this point on, the maintainers will be making this and any other changes to the repos.

Issue #242, opened 2015-01-21
Allow infinite (n) optional version segments


JamesMGreene:
### The Problem

Among the other reasons mentioned in various issues both opened and closed in this repo (#213, #241, #200, etc.), being confined to only a 3-segment version number for public releases is especially bothersome for downstream wrappers (e.g. bootstrap-sass, the Node.js module wrapper for PhantomJS, the Node.js wrapper for Adobe/Apache Flex SDK, etc.) that would benefit greatly from being able to maintain some semblance of being aligned with the upstream library's version number.

For example, with [JamesMGreene/node-flex-sdk](https://github.com/JamesMGreene/node-flex-sdk/), I've wrapped the Adobe/Apache Flex SDK for consumption (i.e. build automation) in Node.js.  The actual Adobe/Apache Flex SDK releases have a 3-segment version number, e.g. `3.0.1`. Ideally, my corresponding version numbers would be identical, plus a wrapper-specific post-build number, so that consumers don't need to learn much about the wrapper itself in order to install the correctly desired version of the upstream library.  I was excited when SemVer `2.0.0-rc1` was posted because it allowed for this exact desire using `+` to indicate the start of a post-build number... but then SemVer `2.0.0-rc2`/`2.0.0` changed `+` to indicate the start of "build metadata" and suggested that it "should" be ignored when determining version precedence. _So..._ back to square one! :black_large_square:

Today, I am choosing to use the pre-release version part for tracking my wrapper's post-build number but, alas, that is incredibly _**un**_semantic, as it would suggest that my wrapper is not on par with the associated upstream version. The result is that someone trying to install `4.6.0` of my Node.js module will end up not getting anything installed at all when they run something like `npm install flex-sdk@4.6.0`. They must instead run something less intuitive like `npm install flex-sdk@^4.6.0-0`.

\*\*\***And please don't just shout "that's a tooling problem!" at this point... NPM is just trying its best to actually adhere to supporting the SemVer 2.0.0 spec correctly for its package ecosystem.**\*\*\*

### Proposed Solution

After looking at #213, which proposes adding a 4th segment to the "core version", my big fear is that we might add a 4th segment in the next major version of the spec (which I am mostly in favor of) but then _that_ would become the new standard for some upstream libraries, and subsequent the downstream wrappers would need a 5th segment, etc. etc. (or a proper post-build versioning scheme that **MUST** be considered when determining version precedence).

[...]

Issue #200, opened 2014-05-20
Nightly builds not supported


mearns:
I really like semver except for specification that all strings following the Major.minor.patch version number indicate a pre-release. I feel that this is inconvenient for, if not incompatible with, rapid development environments with development builds (e.g., nightly snapshot builds).

In my team, we do not always know what the version number of the next release will be, because we don't know what kind of changes will be ready for inclusion when it is time to release. We therefore do not bump the version number until the release. Instead, we add a "post-release" suffix to the version string immediately after the release. For instance, we release version 1.0.0, and then add the string "-dev" to the version string, so that until the next release, we are all working with version 1.0.0-dev. Nightly builds will include this suffix on the version string, indicating that it is a development build.

Conceptually, it is very similar to the idea of a pre-release identifier as you've described it, but the important difference is that its precedence is _higher_ than the version number it attaches to. Again, this is because we don't know if the next release is going to 1.0.1 with a critical bug fix the next day, or 1.1.0 with new features sometime next month.

I would love to see semver updated to support such a feature, though I'm not sure how exactly it could be incorporated without breaking backward compatibility. Perhaps this could be worked into semver ver 3.0, unless you have a suggestion for how semver can handle nightly builds.

JamesMGreene:
By the way, please take a look at this semi-related issue that I opened, discussed, and ended up closing: #242

The discussion may be more specific to my particular use case (wrapping an external dependency) but there's definitely a substantial amount of merit/insight to be gained within, IMHO.  You can ignore the opening discussion and skip ahead to where we discuss build metadata (in the fashion akin to @jeffhandley's notes here for NuGet).

Additionally, I would agree that we really still need post-release versioning added to SemVer while _also_ maintaining the build metadata concept (but ideally with an indicator other than the misappropriated `+` :wink:).

Abdillah:
I have some view on this (nightly build).

When RAD or Nightly Build is applied, I assume that pre-release is vacum; means the development do not need to use alpha, beta, etc. Is it right @mearns?

Then, why don't SemVer gives more freedom on this pre-release "tuple"? I mean instead of specify after a dash is pre-release, SemVer can generalize that it is a tuple representing development stage. It can either pre-release or post-release. But, a project MUST consistent on using it and MUST included explanation on documentation. Moreover, whether pre- or post-release considered as unstable, and both is necessary for API user to review before using or Use at Your Own Risk :tm:.

[...]

PR #519, opened 2019-05-28
[RFC] New post release syntax / change for build metadata syntax

sylr:
A post release syntax is useful for hot fix versions and for people maintaining patch sets on top of official versions that did not make it to the official release (whether is hasn't made the cut or because it has been refused).

According to me the most logical syntax for post-release is to use the `+` sign. As it is already used for build metadata in v2.0.0 I'm suggesting to change build metadata separtor to the `~` sign.

Example: `1.15.3+patchset.1~go1.11`, `1.15.3+patchset.1~go1.12`

Real life example: `linux-4.4.190+rt.187` -> [linux kernel 4.4.190 with RT patch version 187.](https://rt.wiki.kernel.org/index.php/Main_Page)

The downside of this is that would be a breaking change.

EDIT: Change example to use `patchset` instead of `hotfix` which is confusing as hotfix automatically yields minor version increment (one quickly forgets that this RFC aims to enabled non official releases from people who needs to manage their own patch sets).

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -1,4 +1,4 @@
-Semantic Versioning 2.0.0
+Semantic Versioning 3.0.0
 ==============================
 
 Summary
@@ -91,7 +91,7 @@ incompatible changes are introduced to the public API. It MAY also include minor
 and patch level changes. Patch and minor version MUST be reset to 0 when major
 version is incremented.
 
-1. A pre-release version MAY be denoted by appending a hyphen and a
+1. A pre-release version MAY be denoted by appending a hyphen sign `-` and a
 series of dot separated identifiers immediately following the patch
 version. Identifiers MUST comprise only ASCII alphanumerics and hyphen
 [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST
@@ -102,13 +102,21 @@ intended compatibility requirements as denoted by its associated
 normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7,
 1.0.0-x.7.z.92.
 
-1. Build metadata MAY be denoted by appending a plus sign and a series of dot
+1. A post-release version MAY be denoted by appending a plus sign `+` and a
+series of dot separated identifiers immediately following the patch
+version. Identifiers MUST comprise only ASCII alphanumerics and hyphen
+[0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST
+NOT include leading zeroes. Post-release versions have a greater
+precedence than the associated normal version. Examples: 1.0.0+fix.1,
+1.0.0+patchset.7.
+
+1. Build metadata MAY be denoted by appending a tilde sign `~` and a series of dot
 separated identifiers immediately following the patch or pre-release version.
 Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-].
 Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining
 version precedence. Thus two versions that differ only in the build metadata,
-have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700,
-1.0.0-beta+exp.sha.5114f85.
+have the same precedence. Examples: 1.0.0-alpha~001, 1.0.0~20130313144700,
+1.0.0-beta~exp.sha.5114f85, 1.0.0+fix.1~go1.12.
 
 1. Precedence refers to how versions are compared to each other when ordered.
 Precedence MUST be calculated by separating the version into major, minor, patch
@@ -126,15 +134,17 @@ lexically in ASCII sort order. Numeric identifiers always have lower precedence
 than non-numeric identifiers. A larger set of pre-release fields has a higher
 precedence than a smaller set, if all of the preceding identifiers are equal.
 Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta <
-1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
+1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0 < 1.0.0+fix.1.
 
 Backus–Naur Form Grammar for Valid SemVer Versions
 --------------------------------------------------
 
     <valid semver> ::= <version core>
                      | <version core> "-" <pre-release>
-                     | <version core> "+" <build>
-                     | <version core> "-" <pre-release> "+" <build>
+                     | <version core> "+" <post-release>
+                     | <version core> "~" <build>
+                     | <version core> "-" <pre-release> "~" <build>
+                     | <version core> "+" <post-release> "~" <build>
 
     <version core> ::= <major> "." <minor> "." <patch>
 
@@ -146,8 +156,13 @@ Backus–Naur Form Grammar for Valid SemVer Versions
 
     <pre-release> ::= <dot-separated pre-release identifiers>
 
+    <post-release> ::= <dot-separated post-release identifiers>
+
     <dot-separated pre-release identifiers> ::= <pre-release identifier>
                                               | <pre-release identifier> "." <dot-separated pre-release identifiers>
+
+    <dot-separated post-release identifiers> ::= <post-release identifier>
+                                               | <post-release identifier> "." <dot-separated post-release identifiers>
 
     <build> ::= <dot-separated build identifiers>
 
@@ -156,6 +171,9 @@ Backus–Naur Form Grammar for Valid SemVer Versions
 
     <pre-release identifier> ::= <alphanumeric identifier>
                                | <numeric identifier>
+
+    <post-release identifier> ::= <alphanumeric identifier>
+                                | <numeric identifier>
 
     <build identifier> ::= <alphanumeric identifier>
                          | <digits>


FichteFoll:
The current wording does not prevent a release from having both a pre-release and a post-release segment. This seems to only be prohibited in the BNF. However, considering how a pre-release is also a release, it should be allowed to have post-releases of pre-releases (for automated/nightly builds).

Other than that, I like this proposal, as I have voiced years ago in #200.

runeimp:
In FreeBSD (and a few other places) you'll often see code ported to that OS often coming in with say version 1.2.3 to the package system and then being patched shortly thereafter as 1.2.3_1 and then (on rare occasions) repatched with an updated patch to 1.2.3_2, etc. If post release numbers are added to the spec then I'd be happier if it followed "prior art", as it were. This would also be less likely to cause confusion "Why is my semver parser tool caring about the build version now?" or "Why isn't my semver parser tool caring about the new post-release version?", etc.

runeimp:
And personally I'd rather have any code that parses SemVer gack on the addition of a new yet unsupported feature than assume everything is fine because the syntax has new meaning. Following concept of _Path of least surprise_.

That said I also personally greatly prefer the syntax change. `-` and `+` designating pre-release and post-release is MUCH more natural than say `-` and `~` or anything else. And it bugs me that I new if SemVer ever did support post-release versioning that `+` was already used for build metadata. Oh well.

FichteFoll:
> If post release numbers are added to the spec then I'd be happier if it followed "prior art", as it were.

There is a lot of "prior art" available here, however. Arch appends a "post-release" number called `pkgrel` with a hyphen, for example.

The advantage of using `_` over `+` would be that it's less confusing with the current usage of `+` for build metadata (but I believe that to be a poor choice to begin with and this would be a good chance to change it).

jwdonahue:
The spec already covers post releases.  They are every minor and patch release (M.y.z) that come after the major release (M.0.0).  I argue that anyone who feels the need to add additional "post-release" digits to any version scheme, has serious DevOps issues.  I am fine with some version schemes providing a release valve for such poorly planned/executed development environments, but I see no need for SemVer to cover them.

sylr:
@jwdonahue The serious DevOps issue you are referring to is called “contributing to open source”. Something your GitHub account shows you know little about.

jeme:
Please think about this for a minute... What is the exact difference between releasing:

`1.15.3+hotfix.1` vs `1.15.4`

In my opinion you can go to routes with your answer:

A) But my hotfix is **not a full release**, it's just a set of files that I can apply on my software while it's running (Implied by HOT)... 
 - And there you have it, it's not a full release so you shouldn't give it a release version at all, it's fine that you relate it to a release by embedding the release version you intended for it into the hotfix name or whatever you do. But when we see software with a SEMVER version, we expect a fully self-contained package...

B) Nothing, there is no difference
 - And there you have it, just do a release.

----
Now, I am assuming we are in the case of A since we mention hot fixes. And therefore I assume that a so called patch set is also a subset of a full release, however this is something that requires a "reboot" of the software we are patching... (Hot vs Cold)...

Now, when you create a hotfix or a patch set, HOPEFULLY you will also create a full release to go along with it... As in, you don't wan't your new customers (users) to install the latest version followed by a set of hotfixes, no you just wan't to give them the latest...

Which means that we are back to B) there is no difference... Any hotfix or patch set should promt for a full new release, and obviously one that bumps the PATCH version in accordance to the SEMVER spec (as I hope your not applying new features with hotfixes?!??).

Which means we are releasing `1.15.4`, not `1.15.3+hotfix.1`...

Now if you wish to make a hotfix/patch set package that then brings a `1.15.3` up to `1.15.4` thats fine, but I would argue that that has nothing to do with SEMVER.

But that obviously just my opinion.

jwdonahue:
@sylr, it's true, I have been busy getting paid to help create hardware and software used around the world by about a billion people, but I have also been contributing here of late, as well.  But ya, you can find way more examples of poorly run projects in the OSS community than there are good ones, so I guess we can agree on that as well.  

Just an FYI, GitHub stats don't really say much about anybody's life-long contributions to anything but GutHub, which is really not very important in the grand scheme of things.  I've been designing/building hardware for 40 years and made a good living developing software for 30.  

I am just of the opinion that SemVer doesn't need any more digits added to it.  It doesn't need "post releases" either.  The preponderance of requests to add one or more digits, for various purposes, to SemVer is evidence we need a more flexible scheme.  I don't think any one standard is ever going to be good enough, so why not evolve to something like [VesionMeta](http://versionmeta.org) and [VersionSchema](http://versionschema.org)?  A standard you can wield to define your development and release practices that is both human readable and can be processed by machine intelligence, without lock-in to least common denominator, is ultimately going to supplant all the other versioning standards.

BTW: Nuget and, I am pretty sure, NPM (certainly others), already provide support for four digit version numbers.  There's nothing wrong with using alternative schemes.  SemVer is just not a suitable venue to codify those other schemes.

jwdonahue:
@FichteFoll 

>The only problem is the huge breaking change regarding the + symbol, but fortunately it seems quite rarely used.

I can attest that it is in common enough use, that breaking it, will cause a great deal of economic damage.  I know of a build ecosystem with 10K+ users that definitely use it every day (I wrote the tooling for it).  I often use it to simply embed the Git hash(s) that the product was built from.  It's my opinion that redefining the meaning of the plus, minus or period symbols isn't just a breaking change, it would be incredibly stupid to do so, without also adding some kind of [Version Meta](http://versionmeta.org) to the string, in order to disambiguate their semantics.

SemVer 1 is syntactically a subset of SemVer 2, so while there were some subtle rules changes that were indeed breaking between them, in practice, SemVer 2 code generally works just fine when presented with a SemVer 1 string, and I think it may be the case that SemVer 2 is more widely implemented than v1.  Basically, SemVer 2 has become a gold-standard in the community.  I would be okay if v2 turns out to be a subset of v3 syntactically, and preferably semantically, but I do view additions to the SemVer spec, as the equivalent of putting lipstick on a pig. 

If you must dress her up, I think the plus symbol should always be interpreted as the start of the build meta tag, but I also think it should **not** be required for it.  The plus symbol causes no end of problems with URL's, due its special meaning to web servers.  It should be deprecated in favor of some other symbol.  Any new tags or version fields should be denoted by symbol(s) not currently in use, and should be selected to avoid issues with URL's.  

If you must have a band-aide field, why not lead with it.  The semantics are hotfix N for version X.Y.Z yes?  Then why not something like `@1-X.Y.Z`?  If you really need to hotfix `X.Y.Z`, then you've probably already pulled it from your feeds.  Many semver parsers expect strings like `<IgnoredStuff><SemVer>`, so the `@1-` isn't likely to break the SemVer parser.  One drawback is that v2 parsers, won't apply the hot-fix for `X.Y.Z` if they already have it, but they will apply it otherwise, if it matches their dependency spec, and a human looking at a list of semverized package names, is going to see `P@1-` and realize they need to take that hot-fix, so they'll modify their manifest to pull something different.  Well, at least until they update their tooling to v3, which should nearly always apply hot-fixes to existing versions.

Ultimately, we're really talking about a request for tooling changes with this PR.  We get a lot of those here, because folks assume that if they can get the Spec changed, that tool owners will fall in line.  I think the evidence suggests otherwise.

FichteFoll:
I believe there is a misunderstanding in this discussion. This isn't about masking a hotfix release as something that "hotfixes this particular version" because that doesn't make sense in semver's philosophy in the first place. The proper way to release a hotfix is to increase the patch version number.
The idea behind post-release descriminators is to have a more or less common practice for development build artifacts that are specifically *not* explicitly releases.

That said, I do agree with @jwdonahue that this is kind of more of a tooling suggestion or an appendix to semver because it doesn't need a concept of post-releases when it mostly adheres to *users* of a library. The real world begs to differ here and with the wide usage of VCS or CI you might end up with the desire to add a version descriminator to any commit to the `master` branch just so you have a way to name that artifact (probably incrementally) that does rely purely on hashes. I suggest reading up on #200 if you're looking for real world examples.

---

As a compromise, metadata does offer itself to be abused in this way such that you just tack an incrementing number to a previously released version number and optionally the commit hash it was produced from, e.g. `1.2.3+r.45.f8622ab`, which then gets a specific meaning depending on your tooling. Semver by itself claims that the same versions with different metadata are uncomparable, but your tooling might be able to adapt to this if you really need to compare these hacked post-release versions. (In fact, my own semver implementation does this because this wasn't prohibited in semver 2.0.0-rc.2.)

Alas, I believe that a backwards-incompatible change like the proposed one is indeed an almost impossible uphill battle against the legacy of the `+` glyph being used for metadata.

The usage of an underscore `_` glyph, however, would be backwards-compatible and still require tooling to be changed accordingly, but is much more closer in effort to adding an option for including metadata in precedence. The result would be a version like `1.2.3-beta.1_4+windows.22f263b`. I suppose it is debatable whether the underscore stands out enough in a string like this.

jeme:
@FichteFoll Hotfix was explicitly mentioned as part of the Topic in the examples, so I don't think your right in that that was not part of it.

The thing about "CI" not being supported as is, I have to disagree... When you have released your version 2.3.8 and you commit a new story, feature (whatever you do) to your CM which is picked up by your build server, you just started working on the next release...

This could start out being 2.4.0-Beta.x, 2.3.9-Beta.x or whatever your plan states.
Yes, you might not know for sure, but there is no harm in starting with calling it 2.3.9-Beta.x and then later say, "WOOPS we added a feature" and then rename it to 2.4.0-Beta.x and so forth...

Now you COULD also take the more extreme route and just go ahead and adopt CD instead of CI, where you maintain a master branch in a releasable state at all time, under such circumstances it becomes real easy, because at the time you accept a pull-request you KNOW what the next version will be...

I know than many then brings up larger projects where they publish packages and consume them across the build of the entire project, but IMO, if we just treat such internal "sub projects" as if they where just another 3rd party dependency, one that we just happened to build and deliver our self into the bigger picture, then it actually simplifies many things when it comes to Quality Assurance etc. so it's well worth it to do it like that - at least that has been my experience so far...

(And yes, I have even done just that for projects that consisted of 300+ internal packages put together).

grv87:
@FichteFoll, @sylr,  please, don't use `_` in the standard. There are already cases where `+` is not supported, and `_` is used instead. See https://github.com/semver/semver/issues/145#issuecomment-385239489

dcleao:
Support for post-release (similar to maven SNAPSHOT builds) versions is _essential_ to implement a CI system where, after a release (presumably, a public one) it is not yet certain what _type_ of release will follow. Depending on the changes which come to be performed, either a patch, minor or major increment may follow. Assuming either, preventively, may have unwanted consequences. This has been extensively described, and many times, in other related issues of this repository.
It is frequently not feasible or practical to do the type-of-change analysis, and corresponding version bump, on every commit; this is true despite the existence of tooling that tries to automate this task, based on commit messages. Rather, version bumps, of an appropriate type, are made in their own commits, acknowledging the whole set of changes made since the last "public" (non-post) release.

It has many times been called to attention that post-release versions are not the same as the metadata information, and that the latter must not affect the order of precedence...

The discussion and bike-shedding on the actual character that comes to introduce this new section is welcome but, imo, should not defer us from agreeing that this style of versioning is actually needed and that should be part of semantic versioning. Even if this is only used for supporting CIs and _internal_, not public, releases. Developers spend long periods of time with internal, transient releases... We can assume the character is `>` for now, and replace it later with an appropriate one.

Even the discussion of whether the current character that introduces the build metadata section, `+` should be changed so that this new feature can use it, and thus whether a breaking change should be introduced, can be deferred to a later time... after we've settled on the actual semantics and gross syntax of the new post-release section...

Please, guys! This is a feature which is very much needed to support today's versioning styles and CI systems. Not defining this in the semver spec forces everyone to _abuse_ the metadata section and implement custom semver-like parsers and sort rules.

jwdonahue:
>Support for post-release (similar to maven SNAPSHOT builds) versions is essential to implement a CI system where, after a release (presumably, a public one) it is not yet certain what type of release will follow. 

That statement is untrue.  I have personal experience with a half dozen CI systems that use semver. Since only one example to the contrary is required as proof of a false statement, I give you this:

1.0.0
1.0.1-a.CI
...
1.0.93-a.CI
1.1.0-a.flight1
1.2.1-a.CI
etc.

What is needed to support today's versioning _schemes_ and CI systems, is a new and entirely different standard than semver.  Whatever you would change semver to, there's always going to be a huge range of folks who will not adopt it, because it does not meet their needs.  So lets define a standardized way to define your scheme and label your version strings appropriately, such that humans and machines can reason over whatever your semantics are.

My personal preference would be something along the lines of:

0001+checkinId.branchName {CIB}
0002+checkinId.branchName {CIB}
...
1019+checkinId.branchName {CIB}
1020+checkinId.master {CIB}
0.1.0+1020 {Flight}
1021+checkinId.branchName {CIB}
...
1.0.0-beta.1+2561 {SemVer}

The bits between and including the braces are the [{VersionMeta}](http://versionmeta.org) strings that, in this case, identify "well known" [VersionSchema](http://versionschema.org).  It's a project I've been cooking, off and on, for the past couple of years, after spending quite a bit of free time reading every single issue and PR that's been posted here at semver/semver and responding to more than a few of them.

So the way the above history would have played out (ignoring many details) is, a developer sets up a DevOps project with a repo and standard build logic, then starts pushing their changes to a working branch that has CI builds enabled. By **1019+checkinId.branchName {CIB}** the dev(s) are ready to push their first feature for "internal" flight testing and issue a PR to master. Once accepted to master the CIB produces **0.1.0+1020 {Flight}**, which the sales force, executives, secretaries and selected early adopters can get there hands on and start providing feedback.  Many iterations and features later, with many branches feeding into master, we get to **1.0.0-beta.1+2561 {SemVer}**, which is the [VersionMeta](http://versionmeta.org) form of a SemVer version string.

Your developer's relationships with internal/external stake-holders will vary from mine, so the above is just one example of the possibly infinite number of schemes that could be defined by [VersionMeta](http://versionmeta.org), which needs a lot of hard work to get it ready for prime-time.  Note that I am not suggesting the promotion of infinite numbers of schemes, only that a good standard, would allow for them, while providing a range of common, well-known, predefined variations for general use.

ljharb:
Can’t you assume the next one is a patch, until it’s not? Meaning, use 1.0.1 with a prerelease tag. Just because you have a 1.0.1-build.1 doesn’t mean you ever have to have a 1.0.1 - the next release could just as well be 1.1.0 or 2.0.0.

jwdonahue:
@dcleao 

>If I understand correctly what you mean, when 1.0.1-a.CI was created, after 1.0.0, how was it decided that the next (proper) release would be 1.0.1?
>How is it known, by the time of the first commit, that the changes in that commit, and in all future commits until a point when a proper release is decided to be created, will be "patch" level?

It wasn't "decided".  The CI build system simply finds the latest release tag, **1.0.0** in the example, and then bumps the patch number and applies the **-a.CI** tag to it.  That last step says "_I am a simple automation and know nothing of the quality of what I just built_".  As the spec very clearly states, a [prerelease tag indicates that any change may be a breaking change](https://semver.org/#spec-item-9).  That is just one simple example of a "post release" version string using SemVer.

You can also add the smarts to the CI system, to process semantic commit messages to do more than just use patch as a build number.  Either way, any automated system should use the prerelease tag to indicate that anything is actually possible, because short of some really good unit and integration testing, it's often quite hard to get that right.  So you leave it up to the wet-ware to eventually cut a proper release.

Your statement was untrue.  No special "post-release" semantics are required for CI builds.  You must have the prerelease semantics to get the process started, so you have prerelease sequence followed by a release, then rinse and repeat.  You can't get the process started with just a postrelease scheme.  Why complicate the scheme with both, when the one prerelease tag is sufficient?  

Here's the scenario I think you are really hung up about:

1.0.0 
1.0.1-a.vNext
1.0.2-a.vNext
1.0.?-Bugfix.1.0.0
1.0.?-a.vNext

Yes?  Package or interface names can help:

P.1.0.0
P2.1.0.1-a.vNext
P2.1.0.2-a.vNext
P.1.0.1 // Bug fix!
P2.1.0.3-a.vNext

There are other ways to deal with it.

1.0.0
1.0.1-a.vNext.1
1.2.0-a.vNext.2
1.0.2 // Just the bug fix for 1.0.0 minus the vNext work.

But you have to ask yourself, why would you fix a bug in 1.0.0 and not also fix it in the vNext line of work?  

The point here is that postrelease is not needed, prerelease is required and sufficient.  Despite the perennial nature of this topic, it has never gotten any traction precisely for that reason.  

I sometimes think that real driver for those who strongly feel the need to add postrelease or otherwise extend the spec, is they assume that doing so will just automatically convince all of the tooling owners to adopt the new scheme, which just isn't the case.  In fact, the reason the SemVer spec is so stable, is it's pretty much been governed with the needs of those tooling owners in mind.  

There's plenty of tool chains out there that roll their own.  You are welcome to do so as well.  If you do, perhaps you will be kind enough to post and maintain a spec for your peculiar way of doing things?

dcleao:
@jwdonahue:

> Your statement was untrue. No special "post-release" semantics are required for CI builds. You must have the prerelease semantics to get the process started, so you have prerelease sequence followed by a release, then rinse and repeat.

You just keep twisting logic and the semantics to your needs... I don't think that you need to "have the pre-release semantics to get the process started". In your versioning scheme, you create a _pre-release_ of a completely hypothetical and artificial release version — the next patch one —, which may never come exist, just to be able to anchor a "sub-version" number with a may-break-anytime semantics (the pre-release version section) to support future development.

With this, you qualify (may transmit the impression of) your future development as something approaching towards `1.0.1` (a patch release), which, by definition of future development, is almost always not the case...

In the process, you get to occupy the pre-release space of the `1.0.1` version with commits which aren't intended to be patches. When and if you _actually need and intend_ to patch `1.0.0`, separately from your future development branch, you will, with all propriety, develop for `1.0.1-beta.1`, ... versions. Unfortunately, these proper pre-release versions will be all messed up with the previously created pre-release versions...

Many times, you need to patch the last release separately from your current future development — this is because you might have changed the code quite a bit since that release and the fix may need to be made twice...

> You can't get the process started with just a postrelease scheme.

Why not?

> Why complicate the scheme with both, when the one prerelease tag is sufficient?

Simply, because you will be using pre-releases in unintended, dubious ways.
Pre-release and post-release apply to different operations. The semantics of one does not fit the other.

> I sometimes think that real driver for those who strongly feel the need to add postrelease or otherwise extend the spec, is they assume that doing so will just automatically convince all of the tooling owners to adopt the new scheme, which just isn't the case. In fact, the reason the SemVer spec is so stable, is it's pretty much been governed with the needs of those tooling owners in mind.

That's not a driver for me, no. But logic is.

I apologize for my opinion, but by reading issues of the SemVer github repository, over time, I get the impression that its stability is possibly due to the conservationism of its maintainers.

Also, respectfully, I wonder if you opt to only look at the needs of the tool owners that agree with you. I'm almost sure that I've read tool owners and big projects' maintainers, in past issues of this repository, asking for and defending this versioning scheme.

> There's plenty of tool chains out there that roll their own. You are welcome to do so as well. If you do, perhaps you will be kind enough to post and maintain a spec for your peculiar way of doing things?

**Peculiar** way of doing things? You do have a nerve...

remram44:
Something that I don't see mentioned (apologies if I missed it) is the need to support the case where **different people use the same version namespace**. For example, I might be maintaining a Debian or Conda package, or a Docker image from some Python or NPM package. It is very much needed that my own artifacts follow the versioning of the upstream package it contains/wraps, however my packaging might need versioning on its own.

This is the main use case of the current post-release schemes I believe. More than a few of them use a dash for that (Debian might have `2.1.3-6` or `2.1.3-0ubuntu1`, PyPI allows `2.1.3-6` and `2.1.3-post2` ([PEP 440](https://www.python.org/dev/peps/pep-0440/#post-release-separators))). Bumping the patch version is possible for the upstream package, however I can't make my Conda package `2.1.4`, it will conflict with the next upstream patch release. Releasing `2.1.4-0` is possible but very awkward when the upstream version it contains is still `2.1.3`.

Every other issue can be worked around I believe, and I agree with the arguments presented here for *just bumping the patch number*.

I think it is unfortunate that `-0` was normalized in semver as meaning a pre-release and no way was included for post-releases. It pushed competing standards in Linux distros and PyPI which I see as a failure of this spec. At least providing a way to do this in semver (though it is unlikely that those existing package managers will be able to migrate) for example through a new character would go a long way towards strict adoption of semver by new systems.

jwdonahue:
Getting adoption of any standard is equivalent to herding cats.

remram44:
I agree :sweat_smile: however that's different from intentionally ignoring uses cases.

Even if, say, `~` was introduced to mark post-releases, we would have semver exactly backwards to what everybody else is doing (`~` for pre and `-` for post). But at least we'd have the complete set of features!

jwdonahue:
> Even if, say, ~ was introduced to mark post-releases, we would have semver exactly backwards to what everybody else is doing (~ for pre and - for post). But at least we'd have the complete set of features!

Everybody?  If I had some time to spare, I'd go dig up a handful that don't do that.  Perhaps the most common one I've seen in the past is _version_tag_<ws>_qualtiy_tag_.  

Your point is take however.  There do seem to be a large number use cases out there that do it differently.  That's not a strong argument however if the use cases that the SemVer maintainers care about, would be completly broken by the change.

How would you distinguish between SemVer v2 and vN?  If you have a strong argument to alter both syntax and semantics, you must also solve this problem or it's pretty much a non-starter.  Instead, you go put forth the effort to write up the standard you want, give it a non-conflicting name and promote it.

remram44:
Look, I'm not trying to pick a fight. Semver has the merit of existing and that's huge. It's a great solution for those that can use it.

I just think that if there was a way for people who can't use it to have a scheme that looks a little bit compatible to semver, we would all be in a better place. PEP 440 is almost that: semver but with specific words allowed after the dash, with specific semantics; some of it compatible (alpha, beta, rc are pre-releases) and some of it unfortunately not (post, r).

It seems to me that there are ways forward, such as deprecating the use of any word (or most words) after the dash (I think it's safe to assume that if `1.2.3-post.2` occurs today in the wild, it is not meant as a pre-release). If this group has no interest in that, it's fine, I understand, and won't think less of you :wink:

sylr:
I've updated the example, I changed `hotfix` by `patchset` to make it less confusing.

Also added a real life example, the RT patch set for the linux kernel.

jwdonahue:
Is anyone implementing this in any of the packaging tools?  Particularly those that are owned by the maintainers?

runeimp:
I have always felt the need for post release syntax. But I agree with @jwdonahue, et. al. that it is rarely truly needed. The one major exception is patching software for any given OS. My primary example is FreeBSD. Though I'm sure this exists for many other systems (operating systems or otherwise) as well. When someone release a new version of their software through the FreeBSD pkg system it's always something like `app-1.2.3` or `app-1.2.3.4`. But invariably a few days later a patched-for-FreeBSD version is released as `app-1.2.3_1` or `app-1.2.3.4_1` and occasionally a few days or weeks later another patched-for-FreeBSD version is released as `app-1.2.3_2` or `app-1.2.3.4_2`. Both of those patches are specific to making the software in question work properly or better on the specific OS. And, to my knowledge, don't involve the original author of the software at all. These are not original author planned fixes or updates or preparing for the next release. It's just what is needed to make the software in question work properly or optimally on the OS in question. A true post release situation.

endorama:
Hello, I'd like to spend a couple words on this, as I faced the same issue. 

The initial framing of this issue was slightly wrong in my opinion, but still I see the usage in the wild and the need for such a case.

I'd like to vouche for a form of post-relese versioning, because:
- changes to be versioned are not changes to the application sources; the application has a version that is meaningful (i.e. it does not make sense for downstream manteiners to "bump" the version) but an additional information is required
- build metadata _may_ be used but as they are not considered in version comparison (`1.0.0+1` and `1.0.0+2` shall be considered equal by tooling) is not acceptable for this use case
- post-release versioning is visible "in the wild" and takes lots of different forms, while having the same semantic meaning (from distribution packaging to downstream distribution)
- the only available solution today that respects SemVer is using pre-release strings, but tools aligning to the spec would require a version bump to work that would not be compliant

As any standard, the [usual XKCD comic](https://xkcd.com/927/) applies, but this is probably why having it in SemVer would be so valuable: SemVer is widely renowned and it would greatly help in adoption.

In short:
- avoid any breaking change is paramount to drive adoption and avoid distrust in the standard
- this semantic is already in use because is needed, but current implementation usually "breaks" SemVer
- I believe is in the scope and interest of this standard to consider such information as part of the specification, because otherwise the version may lose semantic value

I think there are benefits from "a form of post-release metadata that is considered in version comparison".

jwdonahue:
@sylr, you'll need to resolve conflicts for this PR to be viable.  Otherwise, please close at your earliest possible convenience.

jeme:
@jwdonahue I feel a little bit off about asking people to keep updating their PR (as in this case) before a discussion on the matter has come to a conclusion.

I don't feel like that's very respectful towards @sylr, who has put in the effort, to ask him to keep spending time on this, if in the end it is decided that this is not a good idea and that it won't make it into SemVer, then it will largely just be a waste of his time. (That is purely targeted conflict resolutions, if he or anyone else has any more arguments or changes to bring forth, that is a whole other matter, and that should be welcome)

When a conclusion is made, then it can be put up for grabs to resolve conflicts and incorporate any feedback given and @sylr would perhaps take that on him, but for now, I don't feel like a consensus has been reached here, so I don't think @sylr should spend time on fixing conflicts before we have reached one.

And on that note, being a Contributor, maybe you know who to reach out to in order to move this discussion along. Right now I feel like it's just hanging.

I know the maintainers may be busy and won't have the time to review this, but at the very least they could be notified that this RFC-PR is sort of stale until they get the time, then perhaps they can at least acknowledge that they are aware of the PR, but that they won't have the time to review it for now.

jwdonahue:
@jeme, I think your spot on in your criticism, given that I think this PR should just be tagged as rejected and closed.  

I occasionally make a quick pass through all the PR's and try to get folks to keep them ready, in case the maintainers happen to drop in.  We've managed to get a few of them pushed through that way over the years.  I am an agitator, not a maintainer.  I basically try to get our community to keep the project as clean as possible, because the maintainers do not seem to be very active.

FichteFoll:
Unfortunately, the context of the last comment was lost. I can come to my own conclusion as to what the essence of the proposal was, so to reduce potential confusion by others, I estimate it to be similar to the following: retract and close the PR as "wontfix" due to lack of discussion or input by the maintainers.

While I, too am quite unsatisfied with how the idea (and this PR, which is a result of that) has been open without a comment for 6 years (I consider #200 to be the start of this discussion), just redacting pull requests due to a lack of response does not help in moving forward. At least not untila general direction for the future of semver has been found.

mentalisttraceur:
I would assert that

1. either your "post release" changes the public API in some way, in which case *it is directly against the whole purpose and justification of SemVer* to present the same major, minor, and patch numbers under the same package name, or

2. or your "post release" does not change the public API at all, in which case build metadata is *precisely* the right place for it.

---

In the meantime, I want to confirm that  `+` is very much in use in the world. I just suspect it occurs more behind closed doors in companies than out in the wild. Open source tends to do... less releases where build metadata is the only thing that needs to change.

---

Also! We need to stop treating the build metadata as if it is never ever valid to use it for comparison. No. That's not what it means. You have to interpret it in the context of what SemVer does, which is expose information *about the public API*.

The build metadata "should not be compared" **for the purposes of SemVer semantics**, because it is meant for things that do not effect the public API.

But the build metadata can, is, and *should* be allowed to be compared by all sorts of other tooling that builds additional semantics or functionality **on top of** SemVer. In fact, I would argue that it is not only allowed, but even good practice, to take advantage of the build metadata being compared by higher level packaging tools for version bumps that effect *the packaging* but do not effect the public API of what is being packaged.

I would even go so far as to say that *packaging systems* - things whose job it is to actually store, install, and upgrade packages - *should* be sorting versions by build metadata, and offering those sorted higher as updates.

On the other hand, *SemVer parsers* - things whose job it is to sort versions **only according to what SemVer tells you about the public API** - should *not* compare build metadata.

This gives proper room for multiple "layers" of semantics.

---

Finally, I want everyone to take a step back from their one pet use case and consider the bigger picture, the general abstraction here: what SemVer is really missing is **a way to concatenate multiple SemVer versions together into one version string**.

If I had to argue for any change to SemVer, it would be **either**:

1. Be explicitly permissive enough with what can come after the `+` to enable appending *another whole SemVer version string* after the `+` (then we can have things like `1.2.3+9.8.7-dev.4.{{ git hash }}+0.1.5-foo`).

    -or-

2. Introduce a new character instead of overloading the `+` to allow gluing SemVer strings together like that. Maybe the `~` that was suggested earlier?

Key point: as soon as we permit multiple semvers in one version string, we allow for any and all complex usecases where it matters to have this additional versioning information.

Because that third party patchset? **That should be semver'ed too**.

So `9.8.7+0.0.1` could be used to mean "this is my patchset on 9.8.7, where I have fixed one or more bugs", `9.8.7+1.0.0` could be used to mean "this is my patchset on 9.8.7, where I have made breaking changes", and so on. And then if while your patchset is still in alpha pre-release (`9.8.7+1.0.0-alpha`), I want to release a patchset on top of it to add a feature, so I release `9.8.7+1.0.0-alpha+0.1.0`. And so on. (Replace `+` with `~` or whatever if we go with option 2 above.)

It would be the job of the release maker to document what each of the SemVers applies to, just as it is the job of the release maker to document their public API to begin with.

FichteFoll:
> what SemVer is really missing is **a way to concatenate multiple SemVer versions together into one version string**.

This is more related to #242 than the topic of this PR. I didn't find an open one that relates to this in a quick search.

alexandrtovmach:

mentalisttraceur:
@FichteFoll I agree that it's *more* related to that one, but I assert that it is still very relevant to this issue as well, because if my reasoning is correct then this proposal is redundant with a solution allowing multiple SemVers to be glued together, and the latter is a better way than this proposal to solve the problem this proposal is meant to address.

dcleao:
Showing my support for @mentalisttraceur's proposal of composable versions. I believe it solves all of the post-like cases. 

I wish the maintainers of this specification would stop what I feel as _gatekeeping_ — against **so many**, for **so long**, in the face of **so much proof** — what they think _semver_ is **not** for.

I wish some other more practically and generally adequate versioning standard would come up and get momentum. Then, _semver_ would become irrelevant, and would succumb, embarrassed, forgotten, and there would be no more gatekeeping to be done. In history books, _semver_ would be used as a case study, to teach kids how specification maintainers may become blockers to evolution.

Hope is the last to die. So, I will have hope till I die.

ljharb:
To be clear, the primary job of a spec maintainer - any spec - is exactly that gatekeeping. The proper way to get something new is *to make something new* and have it gain adoption, not to make a thing into something it was never designed to be.

ljharb:
@SamirSaidani yes, that is the nature of time itself: if you change your mind between a "pre" action and the action itself, then the "pre" action doesn't end up saying anything about what version was actually released.

SamirSaidani:
> 
> 
> @SamirSaidani yes, that is the nature of time itself: if you change your mind between a "pre" action and the action itself, then the "pre" action doesn't end up saying anything about what version was actually released.

@ljharb Sure, I'm just saying that there is a way to dramatically increase the probability that the pre-release version will be the actual one or to get rid of the uncertainty, by taking advantage that when you already know that you don't know, you just have to think post-release instead of pre-release. Actually we might even work only with post-release instead of pre-release (kinda "dual" semver).

ljharb:
@SamirSaidani you can't get the probability to 100%, though, so it's still not certainty - so there's not actually a benefit imo.

mentalisttraceur:
@ljharb There is a benefit as soon as you can get that probability at least as high as the probability that the official release SemVer string has a mistake in it.

P.S. In fact the correct way to figure out the combined benefit of different probabilities is *not* to treat the lowest probability as putting a ceiling on the value of any higher probability anywhere else in the whole picture. The correct way is closer to multiplying the value of each possibility by that possibility's probability, and then totaling those weighed values. Being uncertain that I won't die in a freak accident any moment now doesn't really cap the value of getting greater certainty that I'll have enough money to pay the bills next month.

P. P.S. More generally, there is no such thing as 100% probability. 100% probability is a convenient fiction and shorthand among mature consenting engineers and philosophers, delusion otherwise. If someone thinks something in the real world is 100% certain, they are wrong, and it is unfortunate that they might be close enough to correct for practical purposes that reality does not empirically reveal their error to them. So a boolean view where only 100% certainty is valuable is actually equivalent to saying that all possible real knowledge is equally worthless.

PR #547, opened 2020-01-17
Update semver.md

### LINKED ISSUE #490: Clarify points 4 and 5 in Semver 2.0
Many people seem to be having semantic arguments over points 4 and 5. Specifically, when should the API be considered "released for the public"?

```
4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. **The public API SHOULD NOT be considered stable.**

5. **Version 1.0.0 defines the public API.** The way in which the version number is incremented after this release is dependent on this public API and how it changes.
```

I believe the intent is that anything pre 1.0 is considered "beta" software, and is not intended to be used in production environments, or by large organizations that have a requirement of stability. However, people assume that a project can choose to never use the 1.0.0 and get away with any sort of breaking changes, even after releasing the code to the public, documenting the API, and gaining a following of users.

There is also the idea of appending '-beta' to the version number, even if it's pre-1.0, to get around the idea of instability in development code. I believe these suffixes should be allowed, but not to avoid eventual major version changes. Perhaps some rules around suffixes should be in place, such as they are only allowed for major versions, not minor or patches. Perhaps a predefined set of suffixes should be created as well ("-rcN", "-beta", "-alpha", "-customTagXYZ").

To avoid confusion, and semantic arguments regarding when and how to use pre-1.0 and post-1.0 version numbers, this is how I would change the text.

```
4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. It is not recommended for the public should not use 0.y.z version numbers, and the project should release a 1.0.0 version when it is ready for the public to use.

5. **Version 1.0.0 defines the public API.** The way in which the version number is incremented after this release is dependent on this public API and how it changes as defined in the PATCH, MINOR, and MAJOR sections.
```

ajbogh:
Clarified points 4 and 5.

Fixes #490

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -68,12 +68,13 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development. Anything MAY change
-at any time. The public API SHOULD NOT be considered stable.
+1. Major version zero (0.y.z) is for initial development. Anything MAY change 
+at any time. It is not recommended for the public to use 0.y.z version numbers. 
+The project should release a 1.0.0 version when it is ready for the public to use.
 
-1. Version 1.0.0 defines the public API. The way in which the version number
-is incremented after this release is dependent on this public API and how it
-changes.
+1. Version 1.0.0 defines the public API. The way in which the version number 
+is incremented after this release is dependent on this public API and how it 
+changes as defined in the PATCH, MINOR, and MAJOR sections.
 
 1. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards
 compatible bug fixes are introduced. A bug fix is defined as an internal


jwdonahue:
If my product is not "ready for the public to use", why would I burn through SemVer version numbers at all?  If I am not publishing the product, a build number should suffice.

ajbogh:
@jwdonahue I appreciate your comment but I'm not entirely sure how it relates to the PR. Regarding version numbers, you're welcome to use any format but this change only entails the details of items 4 and 5. Other items describe using notations such as "-prerelease" and build numbers. In fact I'm using a package that uses "canary-xyz123" as a version number.

Again, this PR is only to clarify and improve items 4 and 5. There is still some leeway given with the verbiage, such as using the words "It is not recommended for the public". Just because someone does not recommend the usage doesn't mean it's off limits. We just want to point out that certain guarantees in code quality are not provided.

jwdonahue:
@ajbogh wrote:

> We just want to point out that certain guarantees in code quality are not provided.

That's already obvious in the current spec.  It does not require added emphasis.

alexandrtovmach:

AbawZero (approved):

adamralph:
Something feels off with this. I'm not sure that recommendations for and against public use should form part of SemVer. After all, if something is not recommended for public use, why _publish_ it?

alexeyzimarev:
There are quite a few popular projects that avoid going 1.0 for the sake of flexibility to make breaking changes. Kubernetes was on 0.X for a while and used in production by hundreds of thousands. Hugo is another example, it is on 0.81 now and millions use it.

adamralph:
Something feels off with that.

jwdonahue:
> There are quite a few popular projects that avoid going 1.0 for the sake of flexibility to make breaking changes. Kubernetes was on 0.X for a while and used in production by hundreds of thousands. Hugo is another example, it is on 0.81 now and millions use it.

0.x, is not a SemVer version string. 0.x.y is not compliant, because it overloads the minor field, for breaking changes, when the spec clearly says that it's intended use is for non-breaking features. Perpetual 0.y.z, while abhorrent, is allowed by the spec. If you are unable to commit to code reviews and testing, sufficient to engender confidence that your changes aren't breaking, you are free to signal that, with perpetual 0.y.z series of versions.

-----

It is okay for publishers to use non-SemVer compliant versioning schemes. There are dozens of such schemes in common use today, possibly more.

PR #578, opened 2020-06-18
feat: Added linter and workflow; Fixed linter warnings

alexandrtovmach:
Added linter and workflow to validate markdown changes. Used the same list of rules as in [semver/semver.org](https://github.com/semver/semver.org)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,14 @@
+name: Check chagnes
+
+on: [pull_request]
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


alexandrtovmach:
@indirect could you, if this PR looks okay, approve it, merge and then turn on rule to require passing workflow check for PRs to `master`

alexandrtovmach:
@indirect Seems like rule for required CI is not enabled, could you double check?

alexandrtovmach:
aaaand, in this repo no `/check` trigger, only CI linter

indirect:
@alexandrtovmach oh that makes sense! oops 😅 unfortunately, it is not possible to require a check that has never run before, so we have to wait until after it runs:

image

PR #586, opened 2020-06-25
Simplify Backus-Naur notation for identifier character

### LINKED ISSUE #569: Simplify the Backus-Naur notation slightly...
Looking at the BNF for numeric identifier,

```
<numeric identifier> ::= "0"
                       | <positive digit>
                       | <positive digit> <digits>
```
it appears to me that you could simplify it by combining the first 2 alternates.  Since \<digit\> would allow either "0" or \<positive digit\>, shouldn't

```
<numeric identifier> ::= <digit>
                       | <positive digit> <digits>
```
be equivalent?

DannyS712:
Closes #569

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -181,8 +181,7 @@ Backus–Naur Form Grammar for Valid SemVer Versions
                             | <identifier characters> <non-digit>
                             | <identifier characters> <non-digit> <identifier characters>
 
-<numeric identifier> ::= "0"
-                       | <positive digit>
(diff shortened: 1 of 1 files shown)


princesse6688 (approved):

jwdonahue:
I think humans are the intended audience of the spec, so EBNF is preferred, but a valid ABNF would certainly be useful to tool developers. I personally think the BNF's and the FAQ, should reside in separate documents.

PR #598, opened 2020-07-23
Add guidance on non-code file changes

JohnTrammellTarget:


steveklabnik:
While I agree with this principle in general, it also feels out of scope for semver itself, which is about defining a particular API, and then communicating changes to it.

dwdjanika (approved):

Issue #665, opened 2021-02-22
Suggestion: Rename `positive digit` in the EBNF ruleset to `nonzero digit`


dtonhofer:
I suggest to rename the `positive digit` in the EBNF ruleset to `nonzero digit`

Because "positive" is not really an adjective applicable to a "digit": it's a digit, not a number.

So we have the digit "zero" expressed by `"0"` and the other digits, the `nonzero digit`s

thompsbrian:
This is pretty easy to fix and mathematically correct.  @dtonhofer do you mind putting in a fix?

dtonhofer:
Okay, will do it later today (hopefully).

dtonhofer:
Submitted a change request.

Note the following:

- I have not upgrade the version at the top of the page, still at 2.0.0. Technically, should become 2.0.1 (?)
- I have used the name `non-zero digit` rather than `nonzero digit` similar to the existing `non-digit`.It's a matter of taste.

thompsbrian:
Yes, 2.0.1.

dtonhofer:
Changed. Seems to have updated the pull request accordingly.

roniesha1994:
>

dtonhofer:
Good day. What happened to this? It seems to be in limbo. Accept? Reject?

PR #673, opened 2021-02-28
Changed `positive digit` to `non-zero digit`as per issue #665

dtonhofer:
Note:

- I have not upgrade the version at the top of the page, still at 2.0.0. Technically, should become 2.0.1 I suppose.
- I have used the name `non-zero digit` rather than `nonzero digit` similar to the existing `non-digit`.It's a matter of taste.

Fixes https://github.com/semver/semver/issues/665

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -182,8 +182,8 @@ Backus–Naur Form Grammar for Valid SemVer Versions
                             | <identifier characters> <non-digit> <identifier characters>
 
 <numeric identifier> ::= "0"
-                       | <positive digit>
-                       | <positive digit> <digits>
+                       | <non-zero digit>
+                       | <non-zero digit> <digits>
 
 <identifier characters> ::= <identifier character>
                           | <identifier character> <identifier characters>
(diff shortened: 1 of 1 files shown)


thompsbrian (approved):

jwdonahue:
-1 is a non-zero digit, but also not valid SemVer.

jwdonahue (approved):
Your right. This is a good change.

0388699390:
/^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/gm

0388699390:
OTP 0388699390

jwdonahue:
On second thought, "<non-zero digit> ::=" should probably be "<ASCII code points> :=" or some such. Pretty sure there's another suggestion somewhere that the current regex is too lax with '\d' because it includes some non-ASCII range UNICODE "digits".

Mohamed09091:
> Note:
> 
> 
> 
> - I have not upgrade the version at the top of the page, still at 2.0.0. Technically, should become 2.0.1 I suppose.
> 
> - I have used the name `non-zero digit` rather than `nonzero digit` similar to the existing `non-digit`.It's a matter of taste.
> 
> 
> 
> Fixes https://github.com/semver/semver/issues/665

PR #724, opened 2021-07-12
docs: add a regex compatible with Bash

dunglas:
`\d` and capture groups aren't compatible with Bash. I added a compatible regex. I hope it will help.


jwdonahue:
Please add a link to regex101 showing your regex and the test data used for the other regex's.

dunglas:
@jwdonahue done.

jwdonahue (approved):
LGTM, for whatever that's worth.

PR #731, opened 2021-07-25
Create codeql-analysis.yml

MaxFox196:


alexandrtovmach:
Thanks for contribution, but I'm not sure about that, because SemVer repo it's basically one markdown file, and there nothing to check for vulnerabilities.

Issue #778, opened 2021-12-02
Make it clear semver has matured past a "proposal"


aaaidan:
The spec uses words like "we propose" a fair bit, which makes it sound as though it is still a speculative RFC. It seems clear to me that semver is widely accepted and widely used in the software industry (and beyond!)

I propose that we re-word some parts of the spec to be less circumspect, and more directive.

jwdonahue:
It is still a speculative RFC.  There's no one governing board out there that gets to decide what the standard is wrt versioning.  There isn't even a consensus around best practices.  SemVer is one among many that have proven useful.  As for "widely adopted" I would agree, to the extent that there are more than a handful of tools that support it properly. But keep in mind that:

- Many of the most popular of those tools, also support other formats and semantics/workflows as well.  
- Many, if not most of the SemVer adopters, do not agree with each other how it should be correctly applied.
- Some of the most widely distributed software, as in billions of customers, do not use anything at all like SemVer.

I would also point out there are some actual software industry governing boards that have their own well defined standards.

aaaidan:
Thanks @jwdonahue, that all makes a lot of sense. Semvar is just one of many possible strategies and there can be no One True Way for versioning. It’s also not “finished” - and may never be!

For those considering semvar as a versioning strategy (or are simply learning about it), some of the spec wording implies it is still at a very early, and very tentative proposal stage. I think it suggests no real-world agreement or adoption, which is no longer true. 

I suggest updating the spec’s tone to match  semvar’s current maturity. I think this can be done in a way that doesn’t conflict with the points you’ve raised.

jwdonahue:
You are always welcome to present a PR/RFC. Just keep in mind that the tone of the current document has contributed to its extensive adoption. It's a distillation of common practice in versioning. It doesn't over-constrain the problem, but seems to get it just about right.  It is also the inspiration/basis for more formal specifications.

> It’s also not “finished” - and may never be!

If you ask the current maintainers, it's essentially done. They all own packaging tools for a variety of development ecosystems. They all have vested interests in maintaining the status-quo. If you are careful about your proposed new wording, they might accept it. You have to be careful to avoid any prescriptive statements that any one of their tools, doesn't currently abide by. They wish to document only that which is common to all of them. So tonal changes might be accepted. They, me and many others, wish to ensure that we always have the freedom to innovate.

[...]

PR #779, opened 2021-12-02
Replace tentative proposal-style wording

aaaidan:
Strengthen wording to reflect the firm industry support SemVer has gained since the spec was first written.

To address #778

Comments and contributions welcome.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -44,7 +44,7 @@ affecting the API increment the patch version, backwards compatible API
 additions/changes increment the minor version, and backwards incompatible API
 changes increment the major version.
 
-This is called "Semantic Versioning". Under this scheme, version numbers
+Under this scheme, version numbers
 and the way they change convey meaning about the underlying code and what has
 been modified from one version to the next.
 


semver.md
```diff
@@ -31,7 +31,7 @@ specified too loosely, you will inevitably be bitten by version promiscuity
 Dependency hell is where you are when version lock and/or version promiscuity
 prevent you from easily and safely moving your project forward.
 
-As a solution to this problem, we propose a simple set of rules and
+SemVer addresses this with a simple set of rules and
 requirements that dictate how version numbers are assigned and incremented.
 These rules are based on but not necessarily limited to pre-existing
```
jwdonahue, line 36: I think there should be a comma here, while you're at it.  Should be "...based on, but..."

jwdonahue (approved):

jwdonahue:
I like this. I wouldn't call it "strengthening", but they are good changes. Industry support isn't relevant either. The fact is that semantic versioning, in its many forms, is part of the technical lexicon. SemVer as presented here, is one form of semantic versioning, but the proposal here is limited to just this one scheme. I would reword this to indicate that this is one, arguably the most popular, form of semantic versioning (possibly the first scheme to be formally named as such?).

aaaidan:
I just had a look, and I can’t find any evidence of anyone using the phrase “semantic versioning” to describe their non-SemVer versioning strategy. 

Perhaps it’s just buried, for me, by the common usage that “semantic versioning” == SemVer. In itself, I wonder if that indicates we needn’t worry about falsely “claiming” that phrase. 

Realistically, I don’t think it would be reasonable in 2021 and beyond to use the term “semantic versioning” in software circles and not expect the listener to think you’re talking about SemVer. 

If I’m right, I think it’s okay to use these terms more or less interchangeably. 

Thoughts?

jwdonahue:
@aaaidan, WikiPedia mentions several forms of semantic versioning: https://en.wikipedia.org/wiki/Software_versioning#Semantic_versioning

There's always been two kinds of versioning: semantic and numerical, both have histories going back thousands of years.

The "semantic X" craze in the software developers community, was kick started at least as far back as May 2001, in a Scientific American article co-authored by Tim Berners-Lee, James Handler and Ora Lassila titled "The Semantic Web", but they merely defined how **_semantics_**, a field as old as language studies, could be applied to the web, such that machines could process it without having to parse human language. 

According to the Google: Tom Preston-Werner (aka Mojombo) used the phrase "semantic versioning" when he created SemVer back in 2013, but I recall tracing it back to the 2008..2010 range a few years ago, still attributed to Mojombo, but his original repo is no longer available on GitHub.  I think his main contribution was the marriage of a **_common_** syntax **_and_** semantics, with an easy to remember brand "SemVer", at a time when there were a lot of newbies coming on the scene and chaotically reinventing every wheel they could think of.

The history of software versioning is littered with a huge range of semantics and syntaxes, many of which originated in the phenolic board labeling schemes used by the electronics industry, which itself was adopted from common book and journal versioning schemes that date back to the first printing presses. 

SemVer may dominate in the NPM, Nuget, and a handful of other silos, but there are many popular software libraries, operating systems and apps that use other forms of semantic versioning, some of which have more specific labels on them in the [WikiPedia](https://en.wikipedia.org/wiki/Software_versioning) article.

aaaidan:
Thanks for the history, very interesting. 

On reflection, the “this is” sentence risks implying SemVer is the only semantic versioning. As you explain, that’s not true. It’s also doesn’t seem to be critically important, and only reinforces the “we propose” vibe I’m trying to update.

aaaidan:
Not sure whether this PR needs more time for discussion and review. Maybe it can it be merged as is? 
@jwdonahue @JohnTitor 

(I don’t know how the PR lifecycle works for the semver repo. )

jwdonahue:
@aaaidan, while I've had a lot to say about many of the PR's around here, I do not have any control over whether they are accepted. See the [Contributor's Guide](https://github.com/semver/semver/blob/master/CONTRIBUTING.md)

Jpac7-ai (commented):
Change it to a normal place

aaaidan:
Moderators, I think @Jpac7-ai could be a bot, as it’s a new account and has had some posts closed with the spam label.

PR #804, opened 2022-02-17
Remove shall and shall not from the note

spring1843:
The document does NOT use the words "SHALL" or "SHALL NOT" so let's not include them in this clarification.


Infinity190 (approved):

ljharb:
This paragraph is typically copied verbatim whether every term is used or not.

PR #901, opened 2022-12-01
Escape dashes in the `1.0.0+21AF26D3----117B344092BD` example.

lgarron:
When unescaped, the first three dashes are rendered as an em dash at https://semver.org/#spec-item-10

> ... 1.0.0+21AF26D3—-117B344092BD.

Adding the backspaces should prevent this issue.


lgarron:
Screenshot of https://semver.org/#spec-item-10 right now:
Screenshot 2022-11-30 at 19 20 30

JohnTitor (approved):
Thanks!

PR #909, opened 2023-01-16
Fix 'backwards' typo

ttous:
Sources:
- https://wordstylehq.com/backward-vs-backwards
- https://www.quickanddirtytips.com/articles/backward-versus-backwards/
- https://writingexplained.org/backward-or-backwards-difference

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -7,9 +7,9 @@ Summary
 Given a version number MAJOR.MINOR.PATCH, increment the:
 
 1. MAJOR version when you make incompatible API changes
-1. MINOR version when you add functionality in a backwards compatible
+1. MINOR version when you add functionality in a backward compatible
    manner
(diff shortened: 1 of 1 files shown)


ljharb:
I'd hardly call this a "typo" since i've never heard "backwards compatible" causing confusion, and since language is descriptive and not proscriptive, that means it's correct as-is.

That said, this doesn't seem harmful.

mbtools:
Agree with the author. It should be "backward"

JohnTitor (approved):
All the use of "backwards" has "compatible" or "compatibility" and I believe "backwards" is also valid in that context (as [Wikipedia says](https://en.wikipedia.org/wiki/Backward_compatibility)).
I don't have a strong reason to reject though, I think we could merge without RFC as it doesn't have any semantic changes.

Issue #915, opened 2023-02-08
Could we clarifies better in the semver doc PATCH releases and initial development?


camilamacedo86:
By reading the doc https://semver.org/ is clear for me that:

- PATCH releases should not contains api changes and they are reserved to CVEs and bug fixes.
- If the project is not a stable API yet. That means the project will not have MAJOR bumps and because of it we could expected any breaking changes into the MINOR bumps.

However, the text:
> 
Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Brings the idea that new features (api changes) would be fine in PATCH releases with ` Anything MAY change at any time.`

camilamacedo86:
Proposed suggestion to bring clarifications to do the doc: https://github.com/semver/semver/pull/916

ljharb:
What it means is that in `0.y.z`, everything is major, and there *is* no patch. "patch" isn't "the third number", it's "the number that indicates non-additive non-breaking changes", which doesn't exist in a 0.x version.

camilamacedo86:
Hi @ljharb, 

The whole point of we have a convention is to have clear and define criteria/rules to define a contract to address the expectations of those that will consume the API as of those that keep the solution maintained. 

IHMO your argumentations shows inconsistent with the info provided in the doc https://semver.org/. Therefore, the https://semver.org/ shows need to properly cover the stage where the project version is < 1.0.0. Otherwise we would not have this discussion and with very different point of views.

Now, following some comments inline about your replies so far:

> What it means is that in 0.y.z, everything is major, and there is no patch. "patch" isn't "the third number", it's "the number that indicates non-additive non-breaking changes", which doesn't exist in a 0.x version.

That does not seems accurate and very contrary of the points described in this doc. If so, why we have in the doc:

<img width="897" alt="Screenshot 2023-02-09 at 08 03 30" src="https://user-images.githubusercontent.com/7708031/217752657-3cca52ae-82e4-4960-9168-5adea34e8fd3.png">

<img width="800" alt="Screenshot 2023-02-09 at 08 23 57" src="https://user-images.githubusercontent.com/7708031/217756885-387c8c83-35a3-4508-a28a-f4329723d0c5.png">

If you are right here then would not make sense the document say for projects < 1.0.0 to do releases `0.y.z` and should explicit describe to use only `0.y`. 

- Why we would describe to began with 0.1.0 because the changes will be also adding new functionalities and re not only bug fixes? 
[...]

PR #916, opened 2023-02-08
Update: Provide more clarity about the term ANY time for initial development releases

camilamacedo86:
## Description

The text `Anything MAY change at any time` can lead the idea that no rules are required and that PATCH releases can also introduces breaking changes or new functionalities when I understand that the whole idea here is clarifies that no compatible changes will be addressed in MINOR releases since the project has no MAJOR bumps at all. 

So, could we please try to clarifies this one?

Closes: https://github.com/semver/semver/issues/915

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -69,7 +69,7 @@ MUST NOT be modified. Any modifications MUST be released as a new version.
 
 1. Major version zero (0.y.z) is for initial development. Anything MAY change
 at any time. The public API SHOULD NOT be considered stable. Incompatible API changes and 
-new functionality might be introcuded in MINOR releases. PATCH releases still reserved
+new functionality might be introduced in incremental `0.Y.0` releases when `0.Y.Z` releases still reserved
 to backwards compatible bug fixes.
 
 1. Version 1.0.0 defines the public API. The way in which the version number
--- a/semver.md
+++ b/semver.md
@@ -68,7 +68,9 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
 1. Major version zero (0.y.z) is for initial development. Anything MAY change
-at any time. The public API SHOULD NOT be considered stable.
+at any time. The public API SHOULD NOT be considered stable. Incompatible API changes and 
+new functionality might be introcuded in MINOR releases. PATCH releases still reserved
+to backwards compatible bug fixes.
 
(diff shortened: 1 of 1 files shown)


ljharb:
v0 versions have no patch number.

camilamacedo86:
Hi @ljharb,

> v0 versions have no patch number.

So, are you saying that the superset rule is not applied in the initial development. If so, what is the point of work with `0.Y.Z`. Why would we need to have Z releases fi we can distributed to them new features and bracking changes at the same way that we can do with `0.Y` releases?

PATCH version when you make backwards compatible bug fixes

semver.md
```diff
@@ -68,7 +68,9 @@ Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
 1. Major version zero (0.y.z) is for initial development. Anything MAY change
-at any time. The public API SHOULD NOT be considered stable.
+at any time. The public API SHOULD NOT be considered stable. Incompatible API changes and 
+new functionality might be introcuded in MINOR releases. PATCH releases still reserved
```
camilamacedo86, line 72:
```suggestion
new functionality might be introcuded in incremental `0.Z` releases when `0.Y` releases still reserved
```

ljharb:
Ideally you'd start the project with 1.0.0 and avoid v0 versions entirely :-)

camilamacedo86:
Hi @ljharb,

Thank you for your replies. Let's try to discuss this one in : https://github.com/semver/semver/issues/915 where we can keep all thoughts centralized.

jwdonahue:
I have a problem with:

> The public API SHOULD NOT be considered stable.

Okay, so the API can change but breaking changes in the implementation aren't allowed? 

And this is contradictory:

> 0.Y.Z releases still reserved to backwards compatible bug fixes.

Which is it? Stable or not stable?

It's a prerelease, therefore all bets are off with regard to the meaning of any of the fields in the triple. 

This phrase from the [9th clause](https://semver.org/#spec-item-9) makes sense to me:

> A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version.

The "and might not satisfy compatibility requirements..." part of that, tells me what I need to know about "unstable".

So why not just drop [clause 4](https://semver.org/#spec-item-4)?  I think that is what was suggested by @stevelabnik in #915.

In any case, messing around with how prereleases are detected and what they mean is a breaking change.

PR #923, opened 2023-03-09
[spec] document real-world v0 semantics

### LINKED ISSUE #915: Could we clarifies better in the semver doc PATCH releases and initial development?

ljharb:
Per https://github.com/semver/semver/issues/915#issuecomment-1458733037

Fixes #915. Closes #916. Closes #127.

Additionally, I notice the document often, but not consistently, hard-wraps at 80 chars. While I find soft-wrapping immeasurably superior for diff clarity and editability, consistency is also important, so I'd be happy to make a separate PR that updates the entire document to consistently soft-wrap or hard-wrap.

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -1,4 +1,4 @@
-Semantic Versioning 2.0.0
+Semantic Versioning 2.1.0
 ==============================
 
 Summary
@@ -13,6 +13,9 @@ Given a version number MAJOR.MINOR.PATCH, increment the:
 
 Additional labels for pre-release and build metadata are available as extensions
 to the MAJOR.MINOR.PATCH format.
+
+Versions starting with zero are special, and follow the above rules matching
+the pattern `0.MAJOR.MINOR` and `0.0.MAJOR`.
 
 Introduction
 ------------
@@ -59,36 +62,33 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 could be declared in the code itself or exist strictly in documentation.
 However it is done, it SHOULD be precise and comprehensive.
 
-1. A normal version number MUST take the form X.Y.Z where X, Y, and Z are
-non-negative integers, and MUST NOT contain leading zeroes. X is the
-major version, Y is the minor version, and Z is the patch version.
+1. A normal version number MUST take the form 0.X.Y, 0.0.X, or X.Y.Z where
+X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes.
+X is the major version, Y is the minor version, and Z is the patch version.
 Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
 
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development. Anything MAY change
-at any time. The public API SHOULD NOT be considered stable.
-
-1. Version 1.0.0 defines the public API. The way in which the version number
-is incremented after this release is dependent on this public API and how it
-changes.
+1. Version 1.0.0, or the first version published, if lower, defines the public
+API. The way in which the version number is incremented after this release is
+dependent on this public API and how it changes.
 
 1. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backward
 compatible bug fixes are introduced. A bug fix is defined as an internal
 change that fixes incorrect behavior.
 
-1. Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward
+1. Minor version Y (0.x.Y | x.Y.z | x > 0) MUST be incremented if new, backward
 compatible functionality is introduced to the public API. It MUST be
 incremented if any public API functionality is marked as deprecated. It MAY be
 incremented if substantial new functionality or improvements are introduced
 within the private code. It MAY include patch level changes. Patch version
 MUST be reset to 0 when minor version is incremented.
 
-1. Major version X (X.y.z | X > 0) MUST be incremented if any backward
+1. Major version X (0.0.X | 0.X.y | X.y.z) MUST be incremented if any backward
 incompatible changes are introduced to the public API. It MAY also include minor
-and patch level changes. Patch and minor versions MUST be reset to 0 when major
-version is incremented.
+and patch level changes. Patch and minor versions, when present, MUST be reset
+to 0 when the major version is incremented.
 
 1. A pre-release version MAY be denoted by appending a hyphen and a
 series of dot separated identifiers immediately following the patch
@@ -141,7 +141,7 @@ have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700,
       1. A larger set of pre-release fields has a higher precedence than a
          smaller set, if all of the preceding identifiers are equal.
 
-      Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 
+      Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta <
       1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
 
 Backus–Naur Form Grammar for Valid SemVer Versions


steveklabnik:
>  I'd be happy to make a separate PR that updates the entire document to consistently soft-wrap or hard-wrap.

I am unsure how I feel about this personally, we can sort it at some point.

semver.md
```diff
@@ -1,4 +1,4 @@
-Semantic Versioning 2.0.0
+Semantic Versioning 2.1.0
```
steveklabnik, line 1: I think you could argue this means 3.0.0. I dunno. Something to think about. Semver has never really been super consistent with versioning itself. Maybe it's a good time to start that.

semver.md
```diff
@@ -59,36 +62,33 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
...
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development. Anything MAY change
-at any time. The public API SHOULD NOT be considered stable.
-
-1. Version 1.0.0 defines the public API. The way in which the version number
-is incremented after this release is dependent on this public API and how it
-changes.
+1. Version 1.0.0, or the first version published, if lowed, defines the public
```
steveklabnik, line 73:
```suggestion
1. Version 1.0.0, or the first version published, if lower, defines the public
```

steveklabnik (commented):
Thank you so much for this. 

I am unsure how I feel about it right this moment, but I deeply appreciate starting to tackle this. You took a different approach than maybe I was thinking of, I am unsure if that's good or bad! I am unlikely to be able to devote significant time on this until next week.

semver.md
ljharb: yeah, i figured we'd iterate on this line :-) you could argue this is technically *loosening* constraints on v0, or increasing them. Making it v3 would certainly be a safer approach.

ljharb:
I'd love to hear more about what approach you were thinking of. No rush; happy to talk about it next week. I'm also available for a video call if that's easier for you.

jwdonahue:
This is definitely a breaking change.  Your favorite packaging tools are not the only bits that depend on this spec.  If you are going to ensconce just one convention that was traditionally an allowed overlay on top of the spec, I think you should use mine <grin>.

You seem to be trying to say that 0.y.z is no longer a prerelease form on its own, which was not the original intent of versions 1 & 2 of the spec.  Version 2 of the spec and a reading of the FAQ indicate that the 0.y.z form precedes the 1.y.z  initial release form.  I am sure there are many explanations of this in semver/semver issues and probably StackOverflow showing typical version histories of the form:

0.1.0 // implied prerelease as the API is not defined until 1.0.0.
0.1.1 // implied prerelease as anything might break.
0.2.0 // same here.
1.0.0 // First release.

As far as I can recall nobody ever raised any objections to this in the past.  Nothing about the original wording precludes you from adopting your proposed workflow oriented overlays, but no sane external consumer of your product would expect a 0.1.0 version to be complete, build or runtime safe.

jwdonahue:
> That's correct, and has always been the case, since you can have a prerelease of a v0. v0 versions are and have always been full releases.

Which is it?  A prerelease or a full release?  I do agree that you can also add a -pr tag as well, but that doesn't change the fact that a 0.x.y without the -pr tag is in fact a prerelease.

According to the v2 spec:

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Sounds like a prerelease to me.  Initial develop precedes release, so clearly this one form of a prerelease version.  Notice there is no 0.x.y in this spec, the author(s) clearly intended 0.minor.patch.  The mere presence of the 0 in the major field literally means any publication can contain breaking changes so the same rules associated with the pr tag apply.

But then we get 5 (emphasis is mine):

> 5. Version 1.0.0 defines the public API. The way in which the version number is incremented **after this release is dependent on this public API and how it changes**.

So now we have the definition for a release version which follows the initial pre-release definition.  Yet another indication that 0.x.y really is something that comes **before** the first release, clearly making it a prerelease. Hence this proposed change is breaking and I see no language therein which provides a means for implementors to determine which semantics apply to a particular version string.

Then we eventually get to 9:

> 9. A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. ... A pre-release version **indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version**.

This is the second definition of a "prerelease" version and it allows us to go back into development mode without having to drop back to the 0.x.y series.  I should point out that having a dev branch perpetually churning out 0.x.y prereleases that land in a series of release branches:

1.0.0 // After 0.7.23 was tested.
1.1.0 // After 0.8.0 was tested.
2.0.0 // Oh maybe dev's are as far as 333.9999 by now.

Isn't prohibited by the spec. While this may not be sound, there's many other potential permutations on this for organizations that do lots of pre-flighting of prereleases in various A/B scenarios.  Fixing that loop-hole is technically a breaking change, but unlikely to have widespread impact, especially among the current set of package types represented by the maintainers of SemVer.  I've certainly never seen anybody doing this, but there must be tens of thousands of bespoke and ad-hoc implementations in the world today.

I think the spec should stick to the minimum required to define a common syntax and semantics, and leave the tool makers and workflow designers with as much freedom innovate as possible.

ljharb:
A prerelease is only something with the prerelease suffix - it has no other requirements. It may sound like a prerelease, but it’s still not one. That the things that a prerelease indicates may also apply elsewhere doesn’t mean it’s a two-way implication.

jwdonahue:
Looks like a duck, sounds like a duck, it's a duck.

ljharb:
Unfortunately, the map remains not the territory.

jwdonahue:
> Unfortunately, the map remains not the territory.

Ya I was just trying to find Preston's original repo. I remember reading through all of the issues before he handed it off to Phil and they moved the official site here.

ljharb:
@steveklabnik gentle ping, since it's been a few weeks :-)

steveklabnik:
I did not manage to get to this this weekend but I do care about it, my intention is to merge something semantically the same as this, or maybe possibly this, but I need some more time to get my full thoughts down but also didn't want to continue radio silence :)

ljharb:
@steveklabnik another gentle ping, if you want to take some time away from governance drama <3

steveklabnik (approved):
Apologies, once again, for taking forever. Sadly the governance drama doesn't let me get distracted with other work, it just robs me of the desire to do anything :(


Okay. After sitting on this for a while, and even passing it by @indirect a while back, I am comfortable with the way that you changed this.

The only real question to me is the "semver of the spec itself" problem. Which is its own can of worms. I am not sure the proper way to move forward on that, but I *also* don't want to block this on solving a bigger issue. Do you have any thoughts there?

ljharb:
Arguably this is a patch in that it loosens the semantics of v0 versions from “always breaking” to “only sometimes breaking”. You could also interpret it as major because it applies semantics that weren’t there before.

Are there any important ecosystems that don’t follow this already, where this change would break them?

steveklabnik:
> Are there any important ecosystems that don’t follow this already, where this change would break them?

As far as I know, no, it is purely a "bring the spec in line with practice" thing. but i wish we had better test coverage here.

So I think this implies we should do it.

FichteFoll:
It may be reflective of what is used in practice, but semver isn't about "how an API is used". It's about "how an API is defined" and if we consider the rules that semver defines to be the API, then this is definitely a breaking change that requires a major bump.

Example:

With semver 2.0.0, you may include breaking changes in the update from 0.1.0 to 0.1.1. With this change, that would be illegal as any breaking change for a 0.X (X > 0) MUST increment X.

semver.md
```diff
@@ -59,36 +62,33 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
...
-1. Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward
+1. Minor version Y (0.x.Y | x.Y.z | x > 0) MUST be incremented if new, backward
 compatible functionality is introduced to the public API. It MUST be
 incremented if any public API functionality is marked as deprecated. It MAY be
 incremented if substantial new functionality or improvements are introduced
 within the private code. It MAY include patch level changes. Patch version
 MUST be reset to 0 when minor version is incremented.
 
-1. Major version X (X.y.z | X > 0) MUST be incremented if any backward
+1. Major version X (0.0.X | 0.X.y | X.y.z) MUST be incremented if any backward
```
FichteFoll, line 88: In order to prevent overlaps between `0.X.y` and `0.0.X`, a requirement for X > 0 should be added here as well. Note that this would not cover `0.0.0`.

spenserblack: > Note that this would not cover `0.0.0`, which … doesn't really fit into any bucket here anyway. Maybe it should be disallowed?

In practice I think I've only seen this used for reserving a name in a registry, not for any actual release. In other words `0.0.0` releases typically don't even have an API yet to version. So I guess the general usage already assumes that `0.0.0` is invalid?

ljharb: I'm not sure why 0.0.0 would be invalid, it's a valid starting point afaict.

jwdonahue: Other than it being far more likely, that an existing feature could be removed or substantially broken without notice in a 0.y.z publication, I have never treated them any differently than any other prerelease.

I do find it useful that 0.y.z preserves the "feature added" and "fixed bug" semantics, rather than the 0.x.y form that munges patch and feature semantics in y, while giving me zero useful information in x. Our pre-ingestion tests are going to tell me whether something is broken, that's why we run them before we ingest them. We assume breakage, but that doesn't stop us from running whatever tests we can, to continuously track our progress.

Given that this proposed change still allows breakage on the y bump, as consumers, we still have to assume that all prereleases may be broken, so this change in the spec is not an improvement. The fact that we would lose the feature/patch signals, implies a loss of data, hence a step backwards, at least for some of us who pay attention to these sorts of things.

Under the current interpretation of the spec, I know that a passing ingestion test on a y bump is meaningless until I add tests for whatever new features were added or removed.  I also know that a z bump should have fixed at least some of the test case failures. I also know that when a non-prerelease 1.x.y is finally released, it's highly unlikely to be a breaking change from whatever the last 0.x.y was.  Under the current spec, we don't have to have special logic for two classes of prerelease.

I think that the 0.major.minor|patch and 0.0.major schemes simply allows publishers to claim semantic versioning while never really committing to keeping un-signaled breakage to a minimum, or really signaling anything useful at all. It also introduces the weird inconsistency that the value of major can go from x > 0 to x = 0 with each shift to the left.

I also think that we should not consider breaking SemVer without first establishing how to disambiguate between SemVer 2 and all future versions, including patches and etc.

Lxstr:
Even setting aside how much transition confusion this would introduce, I think this PR would cause far too much cognitive load when you consider how millions of people use this concept of semver MAJOR.MINOR.PATCH. Even one second extra is thousands of extra hours spent decoding what semver a package is using, maybe guessing based on it's release date and then trying to understand what changes have occurred. Also, having to communicate about changes in a <1.0.0 version 
becomes too hard and I suspect you'll get people simply not comply with semver, or semi-comply and use 1.0.0 as their starting point. 

Perhaps the problem statement and intent could be clarified, but I suspect the intent is to reduce ambiguity on whether a release is stable or in initial development, and that would be a good idea, but I don't think this is the way to go about it. One idea might be that the Development Status of the product could be more clearly separate from semver altogether. For example Pypi has the underused ability to mark Development Status:

 1 - Planning
 2 - Pre-Alpha
 3 - Alpha
 4 - Beta
 5 - Production/Stable
 6 - Mature
 7 - Inactive

Note decisions regarding this are often not made WITHIN a release, the clearest example would be deprecating a package, but you can apply it to the others. You wouldn't cut a new release to mark something as deprecated, you would flag it as deprecated and archive the repo.

I think it would be very neat if this was used more ubiquitously while semver eventually removed the notion of <1.0.0 being initial development. Think of packages that need to make a breaking change but don't consider themselves beyond a alpha status for their product overall. And of course as you illustrate, there are many sub 1.0.0 products that have become defacto Production/Stable, cutting a new release just to indicate the product is stable seems strange.

ljharb:
@Lxstr this is already how npm (and possibly bundler?) has always operated, and the number of people using npm is quite significant compared to all other users of semver, so I think this would reduce, not increase, confusion.

The intent is to update the spec to match what npm does.

jwdonahue:
> The intent is to update the spec to match what npm does.

That is exactly why this PR should not be accepted. It would impact a large number of valid implementations. There are many other implementations that do not and probably will not conform to how NPM does it. Keep in mind that SemVer predates NPM. There is nothing in the spec that precludes NPM's behaviors, so why change the spec to match what they are doing?

ljharb:
@jwdonahue what highly adopted implementations would it impact?

jwdonahue:
@ljharb Nuget. It took a lot of effort to convince them to support SemVer. And you're looking at this the wrong way around. SemVer is widely adopted and there are many implementations that support it. You're asking for a lot of code changes that provide no discernable benefit.

It's a breaking change that will split the SemVer community. SemVer is likely to out live NPM, so there are likely to be a few more highly adopted implementations down the road, if the spec doesn't become too complex.

jwdonahue:
The SemVer feature I am trying to preserve is that; with the current spec, we can distinguish between bug fixes and new features, when publishing a release or a prerelease. All prerelease publications, including initial development, are potentially breaking.

I haven't used NPM in many years, remind me, does it go:

0.0.1 // patch or feature.
0.1.0 // oops, breaking patch or feature.
0.1.1 // non-breaking patch or feature.
0.2.0 // we had to break it again.
1.0.0 // Ya, that's good

Or does it go from 0.2.0 to 2.0.0?

I seem to recall the first example is how it works. Please correct me if I am wrong. And doesn't NPM drop the build meta tag? I seem to recall it's never been fully SemVer 2 complaint anyway, so why should you care what the SemVer spec says?

It seems to me that NPM employs one of many non-SemVer forms of semantic versioning and they already have their own set of well written documentation.

Lxstr:
NPM recommends to start your packages at 1.0.0 https://docs.npmjs.com/about-semantic-versioning which basically makes this PR at odds with NPM's own recommendation. 

> To help developers who rely on your code, we recommend starting your package version at 1.0.0 ...

Meanwhile Python has no offical recommendation as such, and pretty huge adherence to semver. Easily comparable usage to javascript.

ljharb:
@Lxstr The recommendation is because of the confusion about what the semantics are below v1, because this spec doesn’t describe them - which this PR corrects.

@jwdonahue This PR describes how it works. Basically, in v0.0.X, v0.X.Y, and vX.Y.Z, x is major, y minor, z patch. It’s not that complex.

npm doesn’t let you publish anything with build metadata, but it supports parsing it just fine afaik. That’s orthogonal to this PR.

Since the spec doesn’t contain ranges (that’s #584), this shouldn’t have any impact on users of the spec in the way you describe, but I’d love to hear concrete examples of how it would.

jwdonahue:
@ljharb 

> Since the spec doesn’t contain ranges (that’s https://github.com/semver/semver/pull/584), this shouldn’t have any impact on users of the spec in the way you describe, but I’d love to hear concrete examples of how it would.

Like I've been saying all along, version selectors are implementation defined and always should be. Perhaps the spec should be more explicit about that?

My test lab uses a selector that accepts prereleases, flags any bug fixes that break something new, and adds warnings to the reports, for any feature updates that aren't tagged with test coverage. My pipelines require that the version semantics always differentiate between non-breaking feature changes and bug fixes. The logic is identical for both prerelease modes.

The SemVer spec is, and should always be mute on publish/release processes, and focus only on the simplest semantics and syntax needed to support whatever processes the adopters of SemVer and related tooling might need to have in place.

Your thinking is too java* centric. There are a great many of us who must deal with more than one language, spanning more than one tool chains.

ljharb:
That opinion is reasonable to express about ranges, but I'm not sure why it applies here - again, if you can help me understand concretely how this PR - and only this PR - would negatively impact other adopted ecosystems, such as NuGet, I'd be happy to work on bridging that gap.

jwdonahue:
I did a search on github for SemVer implementations a few years back. There were over several hundred of them at that time. I would estimate that there are perhaps as many as 10K, or more, SemVer implementations being used in production systems around the world today, mostly in the form of scripts. Most of us do not do SemVer the MPM way.

ljharb:
It's clear this change would be a "breaking" change in the spec. That's ok tho, because the long tail of implementations are best served by following what the most popular implementations do - that's how standards work, in practice.

jwdonahue:
@ljharb 

>...but I'm not sure why it applies here...

SemVer semantics influence baseline selector logic. Last few times I read your proposed changes, they effectively modify the semantics of the version triple. It is a breaking change that further restricts the range of conformant selector logic.

ljharb:
Yes, that is the point, and the improvement. This PR would need to be part of a "v3" of the spec, surely.

jwdonahue:
The 0.y.z form, uses the major field to signal that a new version might break something, without requiring the major version bump on each release and without losing the distinction between non-breaking additions and bug fixes. The SemVer spec is built around the mapping between major field indicates breakage, minor field indicates additions and patch indicates bug fixes and the syntax is major.minor.patch.  The meaning of each field is fixed, the values are used for sorting and selecting.

jwdonahue:
@ljharb 

> This PR would need to be part of a "v3" of the spec, surely.

Might as well call it the MPM version of the spec.

ljharb:
*n*pm

(and that sounds like a good thing, given that it's the largest ecosystem by any available metric)

jwdonahue:
His a metric for you:

It's just 1 to maybe 5 in thousands of implementations, hardly dominate.

ljharb:
Number of implementations is irrelevant, anyone can make a hundred new implementations that nobody uses - what matters is the number of users (individually or in aggregate).

jwdonahue:
Not relevant to you perhaps, but to all of us implementers that gleaned the beauty of the fixed syntax and the specific semantics of SemVer, NPM's barely conforming implementation is of little concern to us. All I personally care about here, is preserving the ability to innovate while honing to the semantics of the SemVer spec. I do not intend to modify any code to cover any version of SemVer that does not offer an enrichment of my own products and processes, any more than NPM is likely willing to adopt any breaking spec changes that do not enrich them. Since they already have this particular feature, it's a wash for them if it is not adopted.

One key feature of SemVer is that it is about as tool-chain agnostic as it can be, while remaining relevant. The spec will become irrelevant to all but NPM users, if this change is accepted.

Since you brought up size repeatedly in this thread, I would point out that the most widely adopted and documented versioning scheme is the version quad that Microsoft uses, and other similar schemes. In Microsoft's case, they are gradually moving towards SemVer for many things, but even there, they do not hone to the NPM scheme for anything but Node packages, that I am aware of (FYI: I am not currently employed or contracted at Microsoft).

NPM doesn't give us the full richness available from the current spec and it's selector logic is complex, so we generally use their SemVer tool for that purpose and just assume that all 0.y.z or 0.x.y or whatever might be in them, are breaking changes and we can't know whether any bug fixes are in it, or whether it's a pure feature addition, so... way more attention has to be paid to the release notes, because the NPM tools don't give us a rich enough selector (despite its complexity).

I suppose the later might be automatable eventually, given the recent advances in AI, but not all NPM packages have accurate release notes. The point is, our tooling works as-is and nobody really needs this spec change, not even NPM.

The shifting fields scheme is well documented here and elsewhere, it should not be included in the spec.

conartist6:
This is the kind of change that destroys meaning that did exist. I created meaning when I authored libraries in accordance with the actual guidance given in the spec, and since packages generally declare (in a consistent manner) which version of semver they support, that the majority of distinctions between 0.x and 1.x APIs made in good faith with regard to the 2.0 spec would be cast into doubt, as in the future it would become hard to know whether the author had was aware of the 3.0 specification rendering such distinctions functionally meaningless, especially given that the 3.0 specification must be applied retroactively in order for it to have the benefit of describing what npm has always been doing.

semver.md
```diff
@@ -59,36 +62,33 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
...
 
 1. Once a versioned package has been released, the contents of that version
 MUST NOT be modified. Any modifications MUST be released as a new version.
 
-1. Major version zero (0.y.z) is for initial development. Anything MAY change
-at any time. The public API SHOULD NOT be considered stable.
-
-1. Version 1.0.0 defines the public API. The way in which the version number
-is incremented after this release is dependent on this public API and how it
-changes.
```
conartist6, line 75: I think the other changes are OK but the removal of these sections would, in my view, be a serious error.

ljharb: That's basically the entire point. The distinction of "v1.0.0" isn't valuable or meaningful, and in practice having it has been harmful.

conartist6: It isn't meaningful to you, but I'd argue that there are plenty of people to whom it is a meaningful distinction. To me it's the keystone to the philosophy of semver. The spec hints at this heavily when in the FAQs:

> If even the tiniest backward incompatible changes to the public API require a major version bump, won’t I end up at version 42.0.0 very rapidly?

The way I interpret the answer is "if there are so many APIs in your package that you're realistically in danger of ending up at v42, you really should have split apart multiple packages with more focused responsibilities".

I think that by making 0.42.0 the same thing as 42.0.0 you're removing the key force the agreement exerts that tends to create strong well-factored ecosystems of code, which in my mind was the largest part of the purpose of this document

ljharb: It's just a number, it doesn't matter if you have v42 or v153. Breaking changes should be avoided in every case regardless, but that's an opinion that doesn't belong in this document. (and that doesn't match my understanding of its purpose, which is to safely convey breakage, not to inhibit it)

conartist6: But v153 does inhibit breakage. It usually means that there are several unrelated APIs inside, e.g. all the helper methods from a utility library, or all the components in a whole design system. Most of the time a bump from v152 to v153 means that 49 of the components in a design system had no breaking changes and one of them did. This is not a situation where semver has failed to protect the publisher, but it is a situation where the it has failed to protect the user, who got 49 false positive reports of changes alongside one real change. This is no longer a valuable tool for helping users understand what has *not* changed about their own software.

ljharb: To be clear, I totally agree with your philosophy here - a major bump when a breaking change doesn't affect you is harmful.

This happens constantly in a way that's entirely unrelated to "is it v1 or not", which is why I think this concern is unrelated to that philosophy. The semver spec simply doesn't address nor attempt to address that concern (nor should it, I suspect)

conartist6: If v3 of the spec is breaking, why not just outlaw versions starting with 0? Wouldn't that solve everyone's problems to their satisfaction? It would be clear that any triplet starting with `0` is not semver v3.

The only problem with it is that it isn't a change that can be retroactively applied without opt-in from the users, but I see that as ideal. The great promise of semver is that it allows you to make breaking changes because people who don't want to be broken have the chance to opt out.

conartist6: The benefit is that it requires no changes at all. Semver already requires versions to start with `1`.

ljharb: No, it doesn’t require that.

Outlawing versions that already exist would be incredibly harmful; that should be a nonstarter.

conartist6: I'm saying the spec exactly as written is already correct. These version are not "outlawed", the `breaking.changing.fixing` triplet pattern simply doesn't apply to them. To get that protection you have to opt in by making at least one official release.

conartist6: The fact that semver explicitly doesn't say what semantics apply to any version starting with 0.x is exactly the freedom you need to be able to solve your problem: defining a meaning for those 0.x version numbers that is consistent with the npm ecosystem's historical expectations. The only problem I have here at all is the idea that the two things are one document: that your changes overwrite the underlying spec instead of coexisting with it as they easily could.

conartist6:
So my argument is that there are two versioning schemes: three-digit semver and two-digit semver. They are compatible with each other because both agree on using a triplet, it's just that in the two-digit scheme the first digit is always 0.

For this reason I think it's more likely that users will know them as semver and 0ver. 0ver is currently a joke, but if the 0ver.org proves anything it's that there's plenty of projects for which the two-digit semver model is appropriate and its usage ongoing. The main problem here **is that the site describing this versioning system is a joke, a mean prod at legitimate engineering**.

All the projects listed on that site need the ability to have a semver.org-like document that describes the semantics of triplet-using projects that never intend to go to `1.0.0`.  If we could convince 0ver.org to go up with an actual spec, the primary difference from the semver spec would be the omission of the language about versions `<1.0.0` not being "public" releases, the inclusion of the criteria for major and minor in a two-digit release system, and the inclusion of the same basic parsing rules for which strings constitute valid version numbers (with the addition of the leading-0 rule).

ljharb:
There’s also one-digit, 0.0.x, where every change is breaking. All three are part of semver in the npm ecosystem.

ljharb:
why would it be valuable to have three distinct specs that nobody uses in isolation, and all combine to form what the world actually uses? (as opposed to one spec containing all three forms)

conartist6:
Are you saying this is something you'd agree with as long as the one spec (semver.org) simply defines one-digit, two-digit, and three-digit variants?

ljharb:
I must not be clear on what you're suggesting.

I think the only thing that matters is having a single spec that defines how npm works, which this PR plus the ranges PR would do. I see no value in having any spec for "semver but v1+", or for "v0.x.y" or for "v0.0.x" in isolation.

conartist6:
My confusion is that to be able to say what constitutes a breaking change you have to know what the package author intended. I thought that we were in agreement on that. Each of the versioning systems in real use use a different digit to denote breaking changes. If you say "the spec doesn't differentiate" then it seems to me you're saying "it will never be possible for users to be protected from breaking changes in all real versioning scenarios."

ljharb:
Yes, but this specification doesn't address that - something external to this spec would have to point to it, like package.json in npm, or the equivalent in other ecosystems. I would support that if you pursued it elsewhere, after a spec exists that matches what npm has always done, of course.

conartist6:
npm has always presented the semver spec as its own ideal -- the idea that packages people depend on should be 1.0.0, that was and currently *is* npm's stated ideal. How would that idea survive these changes?

ljharb:
That wasn't always the case, and just because it's the default in `npm init` doesn't mean anyone is prevented or even discouraged from publishing or using v0 packages.

Thus, these changes would have no impact on anything except to bring this specification into alignment with its most widely used implementation.

conartist6:
"It has no impact on anything" is wearing a bit thin when I am describing the impact. You could earn my endorsement though by creating explicit guidance in the language of the specification that describes when it is appropriate to use one digit vs two digit or three digit versions.

ljharb:
To be clear, all of them are three digits - it's just that the first one or two is zero. Adding "should" guidance doesn't really make sense tho when the guidance is "use whatever version you feel like using" - it's not this spec's place to editorialize about the appropriateness of version number choice, as long as breakage and additions are conveyed correctly by transitions.

conartist6:
I said my piece. I believe you should write up a spec that has no editorial intent. This spec has editorial intent, and if it stops having that intent it will need to be replaced with one which does.

ljharb:
This PR does not add editorial intent - it has precisely the same amount as the current spec.

conartist6:
My problem is removing editorial intent. Your starting point is a spec with an opinion (`major.minor.patch`), and at the end you're creating no fewer than three offiically blessed, equally valid ways of doing things and not expressing any preference for one over the other.

Spiker985:
Adding my two cents:

I would argue that the V2 spec explicitly defines 0.Y.Z as developmental, and not to be considered stable in **any** event.

By explicitly defining the rules of 0.Y.Z, I think you're shooting yourself in the foot, as least as written currently.

Just because the npm matcher treats 0.Y.Z as 0.X.Y is not a fault of the spec, but a decision within NPM's implementation.

NPM should follow the V2 as outlined, in that every bump to an 0.Y.Z version, as long as Y or Z have incremented, are potentially breaking changes.

If the author of that software doesn't wish to declare an official release as 1.Y.Z, then it is up to them to relay to their user base that there are no breaking changes - because as per the spec, they should consider every bump as potentially breaking.

Additionally, according to the V2 spec itself, the changes listed in the document would constitute a Major bump as you are explicitly defining (in essence) new ways to handle behavior which would be potentially incompatible with the existing V2 spec.

bvisness:
The changes being proposed in this PR seem very shortsighted and would surely destroy all the shared understanding of semver that the software industry has built up over the past 10+ years.

The current spec explicitly says, and has said since version 1.0:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Redefining 0.y.z to imply any level of stability is straightforwardly at odds with what semver has guaranteed from the start. At the very least it would have to be released as a version 3.0.0, but I believe this change would be a disaster regardless.

Semver is special because the version numbers _mean something_. Unlike a mysterious "version 12345", a semantic version number tells me:

- What level of stability I can expect (0.x.y means fully unstable, >=1.0.0 means API-stable)
- How often breaking changes are released (by the size of the major version)
- How often the library is maintained (by the size of the minor and patch versions)

Under these proposed changes, all this meaning is obliterated. 0.2.0 could now mean "very early, highly unstable, use with caution", or it could mean "highly stable, use with confidence". 0.0.3 could mean "WIP for resume plz hire me", or it could apparently mean "I 100% guarantee backwards compatibility in all respects".

Worse, a _change_ in version number now loses all meaning. How are we supposed to know what a change from 0.0.2 to 0.0.3 means? Worse yet—how are _tools_ supposed to know? If a package bumps from 0.2.0 to 1.0.0, is this to be interpreted as a change from (major 2, minor 0) to (major 1, minor 0, patch 0)? Is such a bump no longer allowed? How then can we even talk about "the" major version or "the minor" version?

---

I understand that this change was driven in part by npm's and cargo's decision to interpret any change immediately to the right of the zero as "incompatible", but any change to the right of a non-zero as "compatible". I understand that this PR is attempting to align the spec with "reality". But I believe this behavior of npm and cargo has been misunderstood.

There is a VERY large difference between "unlikely to break" and "officially a major version". It is indeed the case that a bump of 0.1.0 -> 0.1.1 is less likely to break than a bump of 0.1.0 -> 0.2.0. But it does NOT mean that the library author has chosen to declare the API stable. In fact, even under npm's current behavior, many authors choose to release 0.x versions explicitly to opt out of stability guarantees. **This change would suddenly force package authors to support versions they did not intend to support.**

Practically speaking, it may be reasonable for npm and cargo to choose to interpret some level of "compatibility" even within the 0.x space, in order to minimize real-world breakage. But this should NOT be interpreted as a guarantee of stability from the author.

---

Finally, it is worth noting that [Go modules](https://go.dev/ref/mod#versions) do not follow npm and cargo's convention _at all_. Go considers anything with a major version of 0 or a prerelease tag to be exempt from compatibility restrictions. It will not install multiple versions of a 0.x package at the same time, nor does it interpret any kind of "compatibility" from them.

I strongly oppose this change. Semantic version numbers need to mean something.

jwdonahue:
@ljharb wrote:

> It's clear this change would be a "breaking" change in the spec. That's ok tho, because the long tail of implementations are best served by following what the most popular implementations do - that's how standards work, in practice.

I think historically, that's exactly how they die. Lots of folks adopt a standard and then some big player comes along claiming compliance + some extras, forces those extra changes into the spec, claiming they have a larger following than the many more opposing implementers, just about everybody abandons it, something new comes along, rinse and repeat.

semver.md
```diff
@@ -14,6 +14,9 @@ Given a version number MAJOR.MINOR.PATCH, increment the:
 Additional labels for pre-release and build metadata are available as extensions
 to the MAJOR.MINOR.PATCH format.
 
+Versions starting with zero are special, and follow the above rules matching
+the pattern `0.MAJOR.MINOR` and `0.0.MAJOR`.
```
JReinhold, line 18: I think the wording here (and throughout the document) is somewhat ambiguous in some cases - specifically "moving" the major and minor term between X, Y and Z depending on the amount of leading zeroes.

Questions that can arise from this:

1. Is version `0.3.0` the same version as `3.0.0`, or are they different?
2. Similarly, how about `0.3.1` vs `3.1.0`?
3. Is it valid to go from `0.4.0` to `1.0.0`, or must you instead bump to `5.0.0` directly, React-style?

To be clear, I don't have an opinion on the _answer_ to those questions, all I care about is that the spec is written in such a way that the questions can be answered by just reading the spec.

@dominikg proposed the following alternative terminology in another discussion:

> every change that increased the leftmost non-zero segment can contain a breaking change ( `0.0.1`->`0.0.2` can break, `0.1.0` -> `0.1.1` is stable,  `0.1.2` -> `0.2.0` can break and so on)

This turns it around, to instead of defining a major as the left-most non-zero number, it defines breaking changes instead. With that perspective `0.4.1` would be major 0, minor 4, patch 1, but also say that a change in the minor version would be breaking. IMO this wouldn't cause any of the ambiguity listed above.

I also feel like it follows the spirit of "let the spec follow real usage", because I think the majority of users that sees `0.4.1` would not write that as "major version 4" but rather "major version 0". But this is just a gut feeling, I could be entirely wrong on this.

Anyway, that's just _an idea_ for another way to phrase this whole change, there are many ways to clear up the ambiguity.

ljharb: I think you're right; that's a better framing.

conartist6:
If this change was adopted I would want to see npm start versioning at `0.0.1` by default, because it's the number that gives the most room for growth. In most of my current development I am shipping 20 - 40 versions of a package before I am confident that I have API stability. I obviously don't want the first blessed version to be `40.0.0`, but if I start development at `1.0.0` that is what will happen since there is no room to go anywhere else.

When I am in initial dev I do not wish to claim that any one of my releases is safe to replace with any other, because I have not yet started to make and test those guarantees. This is reflected perfectly in the new system if I release `0.0.1` followed by `0.0.2` ... `0.0.20`. At some point I will start to feel confident that I know the difference between a breaking change and a nonbreaking change, and I will start to have value to me to have that distinction when consuming my own packages, and would "promote" (I think we need a word for this) the package to `0.1.0`.

I might proceed then releasing versions like `0.1.1.`, `0.2.0`, `0.2.1` etc, until a significant number of people are using the tool and it is painful not to be able to distinguish between an addition to the library (purely optional to users) and a fix to a problem with the tool (you probably want to install this and it should as safe as possible). At that point I would promote to `1.0.0` and have reached the final stage of semver maturity as I reached the final stage of maturity in terms of library stability.

Not all libraries would go through all phases of promotion, but I would describe them basically like this:

- `0.0.X` is the "development" phase. Initial-development phase code is the rawest, and a clear marker of volatility to consumers
- `0.X.Y` is the "unstable" phase. It is usable to consumers, as long as consumers are content to keep upgrading. Some libraries will never leave unstable phase, because it is not their goal to reach the kind of stability stability. If you're think you might realistically hit `0.100.0`, park your library in this phase
- `X.Y.Z` is the "stable" phase, used by the most mature libraries, especially those whose ambition is to have very few if any breaking changes over their mature life (such as micropackages). Reaching `1.0.0` is a sign that your core API adheres well to the single responsibility principle.

I would argue that this passes the smell test of being an accurate way of describing (for the most part) what people already do

ljharb:
I would argue that if it's not stable, it shouldn't be published at all, and that there shouldn't be anything special about a low number of majors - if you have 50 breaking changes, you should be on v50.0.0.

conartist6:
Why though? In aviation we have a saying, "the three most useless things to a pilot are the altitude above you, runway behind you and fuel on the ground"

ljharb:
Because the only purpose of publishing something to npm is for people to use it; only stable things should be used; therefore only stable things should be published. If they want to test out something unstable, they can install it from github.

conartist6:
That's the waterfall philosophy in a nutshell. I know you work mostly on mature projects and those very close to the stable core of the Javascript spec, but I work out in the agile boonies. I ship code to figure out what code to build next. Often I ship an unstable library so that I can try integrating it with other parts of my systems so that I can gather the feedback on how the library fulfills its purpose so that I can stabilize the design -- all of this in service of a single goal: reach 1.0.0.

jwdonahue:
The true SemVer way is to start with 0.1.0, then if that first round of features needs a bug fix, you have 0.1.n, until you're ready to release another batch of features in 0.2.0, rinse and repeat. Maybe you get to 0.100.n before you promote to 1.0.0 and get serious about publishing stable versions.  The whole point of the fixed Broken.Feature.Patch pattern is to telegraph to your consumers the kinds of changes you're intending to make, whether you just dropped bug fixes, new features or made breaking changes. The entire 0.Y.Z series are for that unpredictable phase where you might abandon, replace or just accidently break some API's, but you still use major and minor for their intended purpose.

conartist6:
Another piece of prior art on the kind of engineering which embraces instability: https://paulgraham.com/hp.html

I've always loved Paul's description of the work we do, and these bits are particularly relevant:

> What and how should not be kept too separate. You're asking for trouble if you try to decide what to do without understanding how to do it. But hacking can certainly be more than just deciding how to implement some spec. At its best, it's creating the spec-- though it turns out the best way to do that is to implement it. [...]
>
> For example, I was taught in college that one ought to figure out a program completely on paper before even going near a computer. I found that I did not program this way. I found that I liked to program sitting in front of a computer, not a piece of paper. Worse still, instead of patiently writing out a complete program and assuring myself it was correct, I tended to just spew out code that was hopelessly broken, and gradually beat it into shape. Debugging, I was taught, was a kind of final pass where you caught typos and oversights. The way I worked, it seemed like programming consisted of debugging.
>
> For a long time I felt bad about this, just as I once felt bad that I didn't hold my pencil the way they taught me to in elementary school. If I had only looked over at the other makers, the painters or the architects, I would have realized that there was a name for what I was doing: sketching. As far as I can tell, the way they taught me to program in college was all wrong. You should figure out programs as you're writing them, just as writers and painters and architects do.
>
> Realizing this has real implications for software design. It means that a programming language should, above all, be malleable. A programming language is for [thinking](https://paulgraham.com/piraha.html) of programs, not for expressing programs you've already thought of. It should be a pencil, not a pen. Static typing would be a fine idea if people actually did write programs the way they taught me to in college. But that's not how any of the hackers I know write programs. We need a language that lets us scribble and smudge and smear, not a language where you have to sit with a teacup of types balanced on your knee and make polite conversation with a strict old aunt of a compiler.

conartist6:
I'm really sorry if I'm beating a dead horse here, but this latest wave of pressure and commentary is a direct response to hearing from Jordan that there was no question whether this change would land, even in spite of the already significant criticism. Is that true?

steveklabnik:
To be honest, it's been hard to get anyone to care about changing stuff around here.

I still believe the correct path for the specification is to bring it in line with practice. The arguments from theory don't really compel me otherwise, personally. Specifications are a way to bring implementations together, and a specification that does not describe reality is not a useful specification.

bvisness:
My previous message was not just an "argument from theory". I believe that this new framework actually does disagree with practice.

The fundamental value of a semantic version number is not really in specifying how npm resolves version ranges. It's in standardizing how humans communicate about versions.

In the past we just had arbitrary globs of numbers like 1.324.20419-3523. All you can really infer from such a number is that bigger number = newer. But building a package manager obviously requires us to automate some amount of reasoning about versions, and automation requires standardization.

But package managers are garbage in, garbage out. Package managers have to take version numbers at their word, and that means humans have to write good version numbers. The value semver provides is _not_ simply mechanical reasoning about breakage, it's a larger framework for package authors to use to communicate the status of their package.

After all, if you really wanted to design a version system exclusively to satisfy the mechanical needs of package managers, you would only need two numbers: major and minor. The major version increments on breakage, the minor version increments on non-breakage. But semver provides much more than that, and _that is a good thing_—because there is much more that humans want to communicate about their packages.

For example:

- Having separate minor and patch versions allows a package author to indicate the size of a change, even when there are no breaking changes. This is useful for indicating bug fixes.
- Having 0.x be special allows a package author to indicate that a package is early in development. This is useful because it allows package authors to communicate frequent future breakage, which users may want to avoid.
- Having explicit prerelease syntax allows package authors to signal that a new version is hopefully stable but needs testing.

The combination of these makes semver a richer tool for communication. When you see `0.3.14` in the wild, you think "early in development, but frequently updated". When you see `1.0.0` you think "the author says it's ready to use, but I bet there are still bugs". When you see `1.5.12` you think "nice, this package is stable but actively maintained, and likely has some quality-of-life features too". When you see `2.0.0-beta` you think "interesting, maybe I'll try this out before settling on the stable release".

**This new framework destroys all this meaning.** This is what I was trying to communicate before. It reduces semver to this dry, mechanical binary of breaking / non-breaking. And by taking away humans' ability to communicate about versions, you ruin semver, and will cause many more garbage versions to be entered into package managers in the first place.

---

I will say, I sympathize with the posters of #127 and #915. But I think the correct solution is simply to provide some non-normative guidance for how to use the minor and patch versions when on 0.x. For example, perhaps we could give guidance like the following:

> While on major version 0, the minor version SHOULD be incremented for large changes and the patch should SHOULD be incremented for bug fixes, with the understanding that any update still MAY cause breakage since the library is still in initial development.

This would ensure that more package authors implicitly follow the model expected by npm and cargo, without forcing them into undesirable stability guarantees.

jwdonahue:
> Specifications are a way to bring implementations together, and a specification that does not describe reality is not a useful specification.

I agree, and for the most part, that's the effect that the SemVer spec has had. Nuget and hundreds of other implementations are compliant with the current spec, including retention of build meta. In fact, my team (at the time) in Windows build, fought hard to get Nuget brought into compliance and I created a compliant, in-house packaging tool that was still in use by everyone who builds Windows, the last time I looked. I have also reviewed code for dozens of other more or less compliant implementations.

The fact that NPM has never been compliant to the SemVer standard, is no reason to change the spec. Doing so, weakens it as a unifying element.

There will always be some level of non-compliant or compliant++ implementations with any spec, as businesses and developers innovate to optimize profits or performance. You can't write a single spec to cover every variant, but you can have one that is close to the most common denominator. If you get in the habit of modifying the spec to comply with just a handful of those variants, while ignoring the many that are compliant, you lose that connection to reality, as well as the motivation to bring implementations closer together.

jwdonahue:
>> While on major version 0, the minor version SHOULD be incremented for large changes and the patch should SHOULD be incremented for bug fixes, with the understanding that any update still MAY cause breakage since the library is still in initial development.

> This would ensure that more package authors implicitly follow the model expected by npm and cargo, without forcing them into undesirable stability guarantees.

That's definitely inline with the letter and the intent of the current spec, but I don't think that's exactly how NPM does it. I don't recall how cargo handles it, but the current PR basically shifts Major to the right one or two fields and requires you do not break anybody without bumping major. So you lose information about some types of changes being made, in order to signal breakage with highest non-zero field.

It boggles my mind, why it's okay to release a series for breaking changes as 1.x.y-broken, but not as 0.1.y, in NPM, when the spec has clearly said that the later form is for initial development, when any release may or may not break something, intentionally or otherwise.

conartist6:
There are two constituencies here, the 1.0.0-means-nothing folks and the 1.0.0-means-something folks. Both philosophies are valid and have clearly guided people's real decision making. The ask is that there be no change to the spec that excludes one group's needs by exclusively favoring the other interpretation.

If you could tell us that a proposal would have to win the approval of both groups to be accepted it would go a long ways to turn down the temperature and allow us to start working to write new spec language

conartist6:
If the 1.0.0-means-something group isn't even going to have a seat at the table or a say in the decision (and the argument I'm hearing is that we should not), then to keep this productive we should accept the fork and go write ourselves a better spec -- one that clarifies the things that have lacked clarity and is actually backwards compatible with Semver 2.0.0.

steveklabnik:
Sorry, reading the recent comments, it seems like people are suggesting that somehow changes here would make existing libraries incompliant. That's not the case. The spec wouldn't mandate any special "1.0.0" handling or before, it would be like the concept of "Implementation defined behavior."

Right now, these operators are not in the spec *at all*, the goal is to properly document the various ways in which the various implementations decided to implement various aspects of ranges/matchers, so that new libraries have some sort of guidance about which semantics they want to choose, and so that users can understand differences between ecosystems.

steveklabnik:
> The fundamental value of a semantic version number is not really in specifying how npm resolves version ranges. It's in standardizing how humans communicate about versions.

This *is* theory. You are trying to impose one set of semantics on every single library that uses this spec. But *in practice*, that is now how multiple ecosystems use these numbers. Describing what implementations do today *is* standardizing how humans communicate about versions: it's just that some folks don't do it in a way that you think is correct. Trying to bring them into line with what you desire is not practice. It's also just not going to happen on a practical level: npm and cargo have different semantics around what "x.y.z" means as a range, for example, and neither ecosystem is going to make a backwards incompatible change to their large ecosystems based on what the words say here. Nor should they. The specification didn't provide any guidance, so they did what they thought was best. This is true for every implementation, not just these two. The goal is to provide said guidance to reduce proliferation of different semantics in the future. But the "there's different semantics" ship has sailed, years ago.

jwdonahue:
@steveklabnik 

> Right now, these operators are not in the spec at all, the goal is to properly document the various ways in which the various implementations decided to implement various aspects of ranges/matchers, so that new libraries have some sort of guidance about which semantics they want to choose, and so that users can understand differences between ecosystems.

Steve, I think you may have responded on the wrong thread or something. This isn't the PR for range specifiers (#584), it's the discussion on making a breaking change to the way 0.*.* works.  I think you know my views on that, wrt whether it should be part and parcel to the SemVer spec.

steveklabnik:
It's an analogy to a different part of the spec to talk about the overall way in which (if ever) the spec will change: to bring the text on semver.org to reflect how semver is used in practice, by all of the various libraries that use it.

Demonstrably, many ecosystems have production level libraries with x=0. People do not actually rely on x=1 as being special in any specific way, they rely on the differences between various version numbers to determine what has changed between them.

This change conceptually (though I haven't read the text lately) is backwards compatible, in that it's a loosening of what is acceptable. If you want to declare that in your implementation, 1.0.0 has some sort of special meaning, that's fine.

But I'm just repeating the discussion from March 2023. I am not likely to engage with it further until I actually build up the will to actually pull folks together to actually do the work of bringing the spec in line with reality.

conartist6:
> Demonstrably, many ecosystems have production level libraries with x=0. People do not actually rely on x=1 as being special in any specific way, they rely on the differences between various version numbers to determine what has changed between them.

If people who rely on that distinction don't exist, you are then telling me that I don't exist. Please stop doing that.

jwdonahue:
> This change conceptually (though I haven't read the text lately) is backwards compatible,

It is in fact, a breaking change. One that is likely never going to be complied with by most of the implementations that aren't already using the 0.x.y variant. Most of us, already wrote code that supports the 0.y.z form. All of my tools, always treat 0.x.y version strings as likely to be broken, and therefore never automagically sucked into production build systems, though I will look at z bumps to decide whether they are worth the cycles to test them, x bumps always get pushed down the priority queue.

With 0.x.y, the NPM way, I can't tell whether that y bump was intended to be a bug fix or a new feature, so it's not important enough for my attention, until I have some spare cycles available. This means that publishers who use the 0.x.y form of versioning, have a harder time driving uptake of pure security patches.

To be honest, I haven't used node more than a handful of painful times and would never use it voluntarily. I did experiment with the 0.x.y form for about a year on several projects, and will never do it again.

As a publisher, I like being able to advertise "here's a bug fix, and it will probably break some of you, but too bad, this is a 0.y.z patch that is perfectly inline with the intent and letter of the SemVer Spec. As a consumer, I am thinking "oh, it's a bug fix, maybe even a security fix, so I better evaluate whether I should go out of my way to ingest another version of an unstable product.

As an engineer, I try to isolate my systems from the fact that millions of developers do not understand the SemVer spec or frankly, few of the nuances of versioning in general. I think there are too many X.Y version schemes out there, masquerading as SemVer, simply because they appended a "0." to it. If all you really want is Major.Minor, don't bother with SemVer.

conartist6:
As a super informal survey lets look at the home page of npm. Every single one of the top 10 packages is `>1.0.0` and `<20.0.0`.

Of the 36 recently published packages, 26 of them are `>1.0.0` and `<20.0.0`. Of the remaining 10:

- 2 are <1.0.0 but in a way that clearly indicates initial development (the author has bumped most of the other packages in the same namespace to 1.0.0 already)
- 5 are part of unstable libraries (versions `0.94.x` and `0.57.x`)
- 3 are not using semver (version `20240317.0.0`)

That is to say (and I don't know how much more clearly it is possible to say this): at happenstance, 100% of the semver users were adhering to the distinction that `>1.0.0` indicates an API-stable package. That includes the five packages which were clearly production grade but also clearly not API stable and chose to indicate it by not publishing `1.0.0`.

I'm all in favor of changing the spec so that a library that is at `0.94.0` is not considered to be under "initial development." That much clearly is currently out of line with reality and we can offer that minority people better protection under the spec so that there is no question that use of `^` is safe and correct for users of libraries distributed this way.

**However.** The idea that the 1.0.0 does not (in a practical sense) mean anything in our ecosystem is dross which can be disproven by even the most basic direct observation. The hypothesis that there is no meaningful distinction exists predicts that `0.1.x` should be just as common in these results as `1.x.x` and `20.x.x` should be as common as `0.20.0x`. It comes as no surprise to me that the numbers fit my hypothesis better than yours, because I know that even people who never read the semver spec were targeted by name-and-shame campaigns pressuring them to make the distinction (see:  [0ver.org](https://0ver.org), now no longer naming and shaming).

conartist6:
Lets look at how the spec made the distinction without being able to enforce it (I'm paraphrasing):

- It says, "Version 1.0.0 defines the public API"
- It says, "If anyone is using your library in production, it should be `>1.0.0`"
- It says, "If you are worried about your library hitting `50.0.0`, you probably broke the single responsibility principle"
- It says, "If you use a library < 1.0.0, anything can break at any time"

Taken together these parts of the spec language create strong system of incentives for authors to publish `1.0.0`, which combined with the social pressure is why we can observe that the intended effect seems to have been achieved in practice -- especially among the most serious vendors of OSS.

conartist6:
Not only that but semver has very good reason to want people behave in accordance with the single responsibility principle!

If a single library exports 50 functions with different APIs (as my library [iter-tools](https://www.npmjs.com/package/iter-tools) does, for now), then when the major version number changes the consumer knows very little about where in my contracts a breakage has occurred. Semver says, "somehwere in there" and leaves the rest of the work to a changelog. Compare that to having 50 functions in individual packages, each with its own version number. In that scenario a semver breakage is a much more valuable signal. Most people know intuitively that alarms that often go off when they should not are not good alarms, and it is the same with semver breakages. The spec is most valuable to all involved when it creates an incentive structure for authors to apply the system in a fine-grained manner so that consumers can be pointed to an exact location where a contract broke.

In short the single responsibility principle makes it possible to bring contract breakage under accurate computational analysis, fulfilling the foundational goal of semver as I understand it.

bvisness:
@conartist6, I think you've made your stance very clear.

jwdonahue:
@conartist6 

> ...0ver.org...

I love it! Rather than buying into the lie that SemVer is the pinnacle of software versioning, they laid out an identifiable alternative versioning scheme. 0ver and SemVer do not need to be merged, they are perfectly useful independently defined schemes.

conartist6:
The real problem is that right now is that some users of what you might call 0ver are supported by `npm` in a manner that has no blessing from any specification.

I would like to write up a proposed for a semver-0 extension spec in the formal style, which when used would provide an extra guarantee about nonbreakage between patch versions and limit initial development releases to `0.0.x`. Since it only provides stronger guarantees than the full spec, it should be a valid subset (which evidence bears out).

I will write the spec in such a way as to give it no force without opt-in so that it would meet my personal criteria of not making things worse for anyone than they are right now, yet giving people the opportunity to provide clarity and make things better. For example the author of a library at version `0.86.0` is able to say with authority, "I am and have always been releasing in accordance with the semver-0 specification." Npm has worked that way all along, so there is an excellent likelihood in that ecosystem that it would also be true in a practical sense.

ljharb:
npm has defaulted to v1.0.0 in npm init for a very very long time, so the real data to look at is who intentionally chose a v0 version after that default changed.

Most devs don’t read the spec anyways, whether it’s semver or JS or anything else. In the npm ecosystem, people learn how it works by seeing what npm does.

conartist6:
Here it is: a concrete proposal for filling the hole in the specification which currently affects users shipping production code under a `0.x.y` versioning scheme: https://gist.github.com/conartist6/37be2e179d286fe7be276e5abe32ff57

FichteFoll:
Even if the SemVer-0 proposal seems somewhat of a joke, I actually do believe that it solves the issue that npm is having with SemVer 2.0.0 currently: It formalizes the communication of version increments in the major-0 range as an extension of the original SemVer definition with higher priority than SemVer itself for the affected version range without also invalidating all other existing usages of SemVer 2.0.0 (let's be real, there will probably not be another 2.x than 2.0.0). Applications following this scheme could then declare that they are using SemVer & SemVer-0 and npm can document that its default `^` operator behaves according to SemVer & SemVer-0.

As I have already mentioned [last year](https://github.com/semver/semver/pull/923#issuecomment-1634079472) this is a breaking change to SemVer itself and not something than can be taken lightly. I'll refrain from commenting on whether I believe this proposal to be better an improvement for now because I don't want to invest this time, especially when we're not even sure what the impact and cost of releasing a new SemVer major is.

jorenbroekema:
So from what I gather, overall the consensus is that the proposed specifics about version <1.0.0 are fine, but the hold-up is that we can't yet agree on whether this should happen in `SemVer 3.0.0` or in `SemVer 2.0.0 + SemVer-0 1.0.0` where it's an extension spec. 

Main reason for not doing it in SemVer 3.0.0 seems to be that it would be quite a challenge to get other package managers/ecosystems to adopt it because it means that the managers need to start auto-bumping/installing/deduping installs for `z` within `0.y.z` and that package maintainers need to avoid doing breaking changes in `y` bumps for `>= 0.1.0 < 1.0.0` range? That seems like a very reasonable argument to me honestly, this would take time & effort and we'll have the situation where lots of tools will be stuck on SemVer 2.0.0 and we then split the ecosystem between 2.0.0 and 3.0.0 which by itself also creates ambiguity, which is exactly what we want to be reducing with this.

So yeah, I'd really like to see the SemVer-0 extension spec then, because at least it moves us forward and gets rid of the ambiguity around it which is creating all the problems at the moment, and makes reasonable suggestions (as opt-in) for folks wanting to implement SemVer and wondering what the hell to do about v0's handling... One small suggestion from me would be to clarify in the main SemVer spec that by default any v0 is unstable and literally any bump should be considered potentially breaking, and that adoption of the SemVer-0 spec makes this "freedom" slightly less loose.

Spiker985:
> One small suggestion from me would be to **further** clarify in the main SemVer spec that by default any `0.y.z` is unstable and literally any bump should be considered potentially breaking, and that adoption of the SemVer-0 spec makes this "freedom" slightly less loose.

To clarify, that language is already in the existing SemVer 2 definition - the definition that would be changed by the original proposal

jwdonahue:
In #127, @mojobo posted:

https://github.com/semver/semver/pull/127#issuecomment-47832310
https://github.com/semver/semver/pull/127#issuecomment-47835108
https://github.com/semver/semver/pull/127#issuecomment-47964783
https://github.com/semver/semver/pull/152#issuecomment-47991353

He clearly intended that all major version zero publications, should be considered unstable, potentially breaking. So if 0.1.2 breaks something, it's okay. He also hoped that publishers would stabilize their products, such that they could confidently release a 1.0.0 or greater version. Anyone that is currently doing the perpetual 0.x.y thing, should simply release version 1.0.0 and pick up from there, or stop claiming to be SemVer compliant.

PR #955, opened 2023-06-21
Add Swift as  a regex support system

### LINKED ISSUE #954: Add Swift as a Regex Support system

Kyle-Ye:
Close #954


JohnTitor:
This itself looks good, but I don't think it's a good idea to list all the systems/languages here (otherwise we would have to update them endlessly). We can probably tweak the text with "etc.", "for example,", or something like that.

Kyle-Ye:
> This itself looks good, but I don't think it's a good idea to list all the related systems/languages here (otherwise we would have to update them endlessly). We can probably tweak the text with "etc.", "for example,", or something like that.

Update the text with "etc."

Kyle-Ye:
> There's also the testing problem. Adding specific languages implies that someone has tested them in those environments.

Yes, Swift is tested to support this regex. You can check it out with a Swift env locally or by using https://swiftregex.com

Swift has build-in Regex runtime support since Swift 5.7.

https://github.com/apple/swift-evolution/blob/main/proposals/0350-regex-type-overview.md

enezmmt:
Ama çok şükür bu konuda

PR #960, opened 2023-07-16
Fix "dependency hell" description typo

BigTurtle8:
From:
 "Dependency hell is where you are when version lock and/or version promiscuity prevent you from..."
To:
 "Dependency hell is when version lock and/or version promiscuity prevent you from..."


rlidwka:
Hell is a place, not a time. Thus, not a typo.

Issue #981, opened 2023-10-14
modern/extended POSIX-compliant SemVer RegEx (for bash)


har7an:
Hello,

today I was writing an application for a bit of CI-infrastructure of mine that needs to handle semver numbers from applications. Since I prefer to write my CI code in plain bash, I came up with a regex that one can perform in bash to match semver. This differs slightly from the [example for numbered capture groups][1] since POSIX regex (which is what bash uses), as far as I know, has no concept of non-matching capture groups. Here's what I came up with:

```bash
# Regex for a semver digit
D='0|[1-9][0-9]*'
# Regex for a semver pre-release word
PW='[0-9]*[a-zA-Z-][0-9a-zA-Z-]*'
# Regex for a semver build-metadata word
MW='[0-9a-zA-Z-]+'

if [[ "$INPUT" =~ ^($D)\.($D)\.($D)(-(($D|$PW)(\.($D|$PW))*))?(\+($MW(\.$MW)*))?$ ]]; then
    MAJOR="${BASH_REMATCH[1]}"
    MINOR="${BASH_REMATCH[2]:-""}"
    PATCH="${BASH_REMATCH[3]:-""}"
    PRE_RELEASE="${BASH_REMATCH[5]:-""}"
    BUILD_METADATA="${BASH_REMATCH[10]:-""}"
fi
```

The individual parts are captured as follows:
- Capture group 1: Major
- Capture group 2: Minor
- Capture group 3: Patch
- Capture group 5: Prerelease
- Capture group 10: Build metadata

Here is the fully-expanded regex pattern:

```
^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$
```

Maybe this will save someone else an hour of playing with regular expressions. :)

[1]: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

har7an:
I see the website code is hosted in this repo as well. If there's interest, I'll happily turn this into a PR to add a third regular expression at the bottom of the page.

jwdonahue:
You will find a test string here: https://regex101.com/r/vkijKf/1/

[...]

PR #998, opened 2024-01-20
semver: Update FAQ entry on regular expressions

har7an:
with a third variant that works without non-capturing groups and is compatible with POSIX regex (i.e. for use in bash). This originates from an issue I opened earlier, found here: https://github.com/semver/semver/issues/981

There is a [small repo][2] which contains the terse form of the regex along with a more human-readable one and a bash script that demonstrates and runs the regex against the [set of test strings from regex101][1]. I also went and [created a snippet for this regex on regex101][4].

I only just noticed that there is a [long-standing open PR][3] which adds exactly the same regex (but with less explanation) to the FAQ.

[1]: https://regex101.com/r/vkijKf/1/
[2]: https://github.com/har7an/bash-semver-regex
[3]: https://github.com/semver/semver/pull/724
[4]: https://regex101.com/r/7XA2jr/1

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -336,7 +336,7 @@ name and the semantic version is "1.2.3".
 
 ### Is there a suggested regular expression (RegEx) to check a SemVer string?
 
-There are two. One with named groups for those systems that support them
+There are three. One with named groups for those systems that support them
 (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python
 and Go).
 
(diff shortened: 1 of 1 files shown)


jwdonahue:
> I only just noticed that there is a https://github.com/semver/semver/pull/724 which adds exactly the same regex (but with less explanation) to the FAQ.

Doh!  I forgot all about that, sorry.

har7an:
@jwdonahue Any news on this?

jwdonahue:
@har7an 

I think we're waiting on the maintainers.

PR #1003, opened 2024-01-29
Grammar: "judgment" → "judgement"

Doomster14:


ljharb:
In US english, `judgment` is the only correct spelling, and in UK english, `judgment` is used in legal contexts, so i think it's better off as it is.

JohnTitor:
Thanks for the PR but I'm going to close this in favor of the above comment. Thank you @ljharb for clarifying!

Issue #1004, opened 2024-02-02
MAJOR should not be incremented unless backward incompatible changes exist


carlwiedemann:
This is a proposal likely not novel. In searching previous issues & PRs I didn't see anything about it.

I'd like to propose that a major increment shall not be made unless backward incompatible changes are introduced.

The language of specification item 8 would change. Something like this (bold text indicates something akin to what I propose):

*8. Major version X (X.y.z | X > 0) MUST be incremented if any backward incompatible changes are introduced to the public API. It MAY also include minor and patch level changes. **It SHALL NOT be incremented unless backward incompatible changes are introduced to the public API.** Patch and minor versions MUST be reset to 0 when major version is incremented.*

I am hoping (perhaps in futility) this dissuades some of the capriciousness that I have seen with major increments. See https://github.com/mbleigh/acts-as-taggable-on/issues/1063#issuecomment-1924464963 for an example if you are keen.

Thoughts? Apologies in advance if this has been discussed & resolved in the past an I have simply missed it. Thanks!

ljharb:
Duplicate of #946.

There's nothing wrong with incrementing the major and having no breaks out of an abundance of caution - quite often a given user won't have any breakage even if the release *does* have breaking changes anyways.

carlwiedemann:
Thanks for linking the other issue, I appreciate it.

I generally agree with the opinion about "over-advertising" breakage vs "under-advertising" breakage.

That said, I'm still curious about what downsides making a change to item 8 would have. Perhaps I would instead propose using "should not" instead of "shall not" to acknowledge [your comment](https://github.com/semver/semver/issues/946#issuecomment-1553261056) that "it violates the spirit of semver, and is a gross and icky thing to do".

Perhaps I could put phrase it this way: In the world of FOSS, inappropriate versioning is going to happen *de facto*, whatever the specification decrees -- robust as it is, we'll still see (for lack of a better word) _careless_ manintainers create breaking changes in minor & patch versions.

But conversely (for lack of a better word) _careful_ maintainers may want to rely on a firmer specification than what exists today, to perhaps resolve debates whether a given change requires a major version bump. In my view, adding a "should not" (instead my previous suggestion of "shall not") could potentially help this debate move along, and avoid some of the "gross and icky" things we all have seen.

[...]

PR #1005, opened 2024-02-03
Issue 1004: MAJOR should not be incremented unless backward incompatible

carlwiedemann:
See https://github.com/semver/semver/issues/1004 for discussion

This is likely an RFC, but I don't believe I have the permissions to add that label myself. Apologies if I missed something about the contribution guidelines.


carlwiedemann:
🦗

PR #1017, opened 2024-04-23
semver.md: add mermaid diagram

lskatz:
I'm wondering if anyone wants to help make a mermaid diagram for this? The SVG is fine, but now that we have mermaid fairly standardized, why not have it actually written up? I started with the easy part but it gets complicated fast.


gwynne:
Honestly, my only complaint would be that Mermaid's rendering is generally kinda awful-looking unless one spends considerable time tweaking it 😕

Eusinho1985:
> I'm wondering if anyone wants to help make a mermaid diagram for this? The SVG is fine, but now that we have mermaid fairly standardized, why not have it actually written up? I started with the easy part but it gets complicated fast.

PR #1218, opened 2026-05-26
Address myth that 1.0.0 is feature complete, fixes #734

### LINKED ISSUE #734: The myth that releasing version 1.0.0 implies any type of long term support
In one widely-used project we are discussing whether to release version 1.0.0.

Some opposing viewpoints cite that version 1.0.0 has some implicit representation of quality or completeness. Or that making a major version release implies that there will be long term (permanent?) support for that release. 

I think the 1.0.0 question of SemVer is one the most important questions. So I love to hear any of your experience on these topics. 

An instance of this kind of discussion: https://github.com/ethereum/solidity/issues/11690

fulldecent:


ljharb (commented):
LGTM!

zafarkhaja:
This whole FAQ item is already problematic, and the change just makes it even more confusing.

> If your software is **being used in production**, it should probably already be 1.0.0.
> ...
> Releasing 1.0.0 is **not a guarantee of quality** or completeness of your software.

Perhaps, some quality? Given that it's being used in production...

---

Aside from this, the spec can't possibly address every baseless myth out there and specify all the things version 1.0.0 doesn't guarantee. Instead, we should focus on clear and precise wording, so we don't have to explain 1 spec point with 3-4 FAQ items and still end up with myths like this.

For instance, the spec _seems_ to suggest that the difference between `0.y.z` and `1.0.0` is in stability of public API. But how do we define API stability? What does "Anything MAY change at any time" mean? What does "Version 1.0.0 defines the public API" mean? These points should be addressed first.

fulldecent:
Here is a concrete, large project, in production that "supports SemVer" and will never release version 1.0.0 because they are scared of mythical things, and because they do not understand SemVer.

https://github.com/argotorg/solidity/issues/11690

I hope we can use this as a case study in finding how to directly improve the FAQ here.

---

I would also like to see a rewritten FAQ. Please propose a separate PR if you would like to do that. I notice that you do not have a "contributor" or "member" badge next to your name, and neither do I, so my instinct is to use a very small change first to address this specific problem. And then if that is accepted and merged, or if a project member invites me to do so, then go and be more ambitious with a larger change.

Issue #1148, opened 2025-09-07
✨ Proposal: Clarify SemVer handling of build metadata and equality


erify-world:
Context

While SemVer 2.0.0 specifies that build metadata (+...) does not affect precedence, there’s ambiguity in how equality should be interpreted across implementations.

For example:
	•	1.0.0+build.1 vs. 1.0.0+exp.sha.5114f85
	•	Should these be considered equal (since build metadata is ignored for precedence), or distinct (since the metadata differs)?

⸻

Problem
	•	Different tooling ecosystems (npm, Maven, Cargo, etc.) treat equality and metadata differently.
	•	This creates interoperability issues when comparing, caching, or distributing artifacts.

⸻

Proposal
	1.	Clarify Equality Rule:
Explicitly state whether two versions differing only by build metadata should be considered equal.
	2.	Reference Examples:
Add normative examples to the specification, e.g.:
	•	1.0.0+abc == 1.0.0+xyz (if build metadata is ignored)
	•	Or explicitly declare them distinct (if metadata counts).
	3.	Impact Assessment:
	•	How will this affect existing package managers?
	•	Should metadata equality be left implementation-defined, or standardized?

⸻

Motivation
	•	Ensures cross-tool consistency.
	•	Reduces ambiguous comparisons across ecosystems.
	•	Provides clear guidance for future spec versions (e.g., SemVer 3.0).

⸻

🔗 Related Issues:
	•	[Equality and build metadata](https://github.com/semver/semver/issues/1134)
	•	[Proposed v2.0.1](https://github.com/semver/semver/issues/1120)

ljharb:
[The spec](https://semver.org/#spec-item-10) says
> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

Comparing two versions for equality is not for determining version precedence.

norman-abramovitz:
I cannot if we are agreeing or not.  The spec allows for version comparison only.  The prerelease section is lower precedence. 
 
The spec intentionally does not define whole string comparison.  Just like adding a v in front of the major number is not a semver string.

Whole string comparison is out of scope for this version of the spec.

ljharb:
Right - this issue is asking about whole string comparison of the semver string - in which case, regardless of the spec's silence on the matter, I'd say the only possible interpretation is that `1.0.0+build.1` is not equal to `1.0.0+exp.sha.5114f85`

smilly-boy:
> Context
> 
[...]

Issue #1120, opened 2025-05-19
Proposed v2.0.1: BREAKING.ADDING.FIXING


Incognito:
There's a lot of misuse or misunderstanding of SemVer in the wild. Common example we all experience: people break the public API but do not increment MAJOR.

I think a big part of this is that yes of course people are not reading the documentation, but we could dig a pit of success for these people to fall into instead of trying to teach them all how to use it correctly. Presently, the names of the numbers is closer in semantics to importance assigned to a number instead of semantics related to the change itself.

I propose that we rename these numbers in the docs to something semantically aligned with what an increment in the number does instead of what position it holds in the sequence.

There's a lot of room for wishy washy interpretations of what "major" and "minor" mean (especially for people who don't read the spec in detail, or lack experience to learn from someone who understands it). There is much less room for ad-hoc interpretation when they have something like "Breaking.Adding.Fixing" instead.

"This is the number I increase when it breaks someone who updates the package version".
"This is the number I increase when it fixes a bug but doesn't add a feature or break the API".
Etc.

Perhaps you disagree with the specific words in my proposal, if so, do you disagree with the general concept of renaming the concept in the docs too?

This change should be fully backwards compatible with the package ecosystem and (as I see it) does not "add a new feature".

If agreed I can submit the change request. I think this would make a gentle nudge on the global package ecosystem towards a slightly better better state.

dunyakirkali:
I like this idea

zafarkhaja:
Sure, better names won't hurt and might go a long way, but I believe coming up with something comprehensive will prove to be the biggest challenge. I know the proposed naming is just an example but to highlight the issue, under this scheme how does one identify the number to increment, let's say, for "deprecating" changes?

Just as an idea, what if we approach this from the opposite side, meaning not in terms of API changes but in terms of changes to client code that the API changes will entail should you decide to upgrade the dependency?

tanadedtk-eng:
Ok

ghost:
https://github.com/semver/semver/issues/1120#issuecomment-3335947551

jwdonahue:
This has come up in the past:

- Rename "Major" to Breaking and "Minor" to Feature #411 
- The terms MAJOR.MINOR.PATCH are not semantic #625

Here's what the original author had to say on the topic:

https://github.com/semver/semver/issues/411#issuecomment-347050750

[...]

PR #1230, opened 2026-07-19
Bump version to 2.0.1 and Clarify build metadata rules

gistya:
This PR addresses the issue where the existing item 10 constrains precedence but is silent on identity. This led to implementations filling that silence by inferring identity from precedence, which is why `exact("1.0.0+Debug")` gets resolved to `1.0.0` in the wild (even when a `1.0.0+Debug` is available). 

To resolve this, the new item 12 names identity as a separate relation and forbids deriving it from precedence.

Some tools which discard metadata and fail to match versions specified in dependency requirements will be out of conformance, but it's not a breaking change because they were never conforming to begin with. They were relying on the spec's silence about identity, not on any guarantee it made. Item 10 already constrained precedence to ordering and said nothing about whether two version strings can denote the same version. Treating equal precedence as license to substitute one version for another was always an inference, never a rule. Item 10's existing wording "two versions that differ only in the build metadata" implies that build identifiers refer to unique versions, so this change simply states explicitly the rule that was already implied. No conforming implementation needs to change.

Related comments: 

https://github.com/semver/semver/issues/1148#issuecomment-3263484442 (and below)

Existing 2.0.1 PR: 

https://github.com/semver/semver/issues/1120

### DIFF
--- a/semver.md
+++ b/semver.md
@@ -1,4 +1,4 @@
-Semantic Versioning 2.0.0
+Semantic Versioning 2.0.1
 ==============================
 
 Summary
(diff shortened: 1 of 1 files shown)


semver.md
```diff
@@ -127,8 +128,8 @@ have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700,
       Example: 1.0.0-alpha < 1.0.0.
 
    1. Precedence for two pre-release versions with the same major, minor, and
-      patch version MUST be determined by comparing each dot separated identifier
-      from left to right until a difference is found as follows:
+      patch version MUST be determined by comparing each dot separated
+      identifier from left to right until a difference is found as follows:
```
gistya, line 132: This change cleans up an 80-width violation

